PR #2, opened 2021-09-03
docs - add info on working group members and meetings

e-lo:
Added:
- working group members
- meetings
  - meeting presentations
  - forthcoming meetings
- links to mdip
- cute bus as logo

PR #3, opened 2021-09-03
Fix devcontainer configuration

thekaveman:
Reorganize and fix the startup sequence that was preventing the site from launching locally within the devcontainer.


thekaveman:
Will redo once #2 is merged.

Issue #60, opened 2024-02-12
new file: run_events.txt


skyqrose:
This is a combination of #51, #52, and #54, updated based on discussion and assuming some form of #55 is accepted.

I changed the name from #51, I now propose `run_events.txt` not `runs.txt` because the file has one row per event, not one row per run, and it [merges in the existing `run_events.txt` table](https://github.com/cal-itp/operational-data-standard/issues/51#issuecomment-1861883775).

This is a concrete proposal, meant to close issues rather than open them. It's intended to be able to be accepted as is, without any open questions or TODOs. Though, of course, I expect there to be discussion and minor changes. I will edit the proposal with any changes.

## Summary:

- A new file `run_events.txt`, which lists all of a run's trips, deadheads, and events.
- Compared to previous versions, now working on a trip/deadhead is now just a normal event, but where the `trip_id` field is set.
- This replaces the existing `runs_pieces.txt` and `run_events.txt` files. Those files would be removed. (`deadheads.txt` would also be removed by #55.)

## Full Documentation:

Primary key: `run_event_id`

<table>
<tr>
<th>Field name</th>
<th>Type</th>
<th>Required</th>
<th>Description</th>
<th>(proposal comment)</th>
</tr>
<tr>
<td><code>run_event_id</code></td>
<td>ID</td>
<td>Required</td>
<td>Primary Key</td>
<td>Added in this version because no combination of other fields are suitable for the primary key.</td>
<tr>
<td><code>service_id</code></td>
<td>ID referencing <code>calendar.service_id</code></td>
<td>Required</td>
<td>Identifies a set of dates when the run is scheduled to take place.</td>
<td></td>
</tr>
<tr>
<td><code>run_id</code></td>
<td>ID</td>
<td>Required</td>
<td></td>
<td></td>
</tr>
<tr>
<td><code>piece_id</code></td>
<td>ID</td>
<td>Optional</td>
<td>
Identifies the piece within the run that the event takes place.

May be blank if the event takes place out of a piece, like a break, or if the agency does not use piece ids.

</td>
<td></td>
</tr>
<tr>
<td><code>block_id</code></td>
<td>ID referencing <code>trips.block_id</code></td>
<td>Optional</td>
[...]

Issue #55, opened 2024-01-03
Supplementing Public GTFS data (trips, stops, stop_times, routes, etc.) within ODS


jeffkessler-keolis:
# Context

The ODS data model is based on the concept of supplementing public GTFS data with internal operational data, capable of together modeling the entire network.

This, in theory, works well: schedule information for public trips is released publicly to customers, and ODS contains all of the non-public information.

# Problem

This begins to break down when there becomes "supplemental" information regarding public trips that some operators release publicly, but others do not. For example:

- **Blocks**: By relying on blocks in trips.txt for ODS's block information, this would require operators to release block information publicly, which is untenable for most rail operations. A big factor influencing this is the frequency of cycle swaps in terminals and the decoupling of crews and equipment, meaning consumers using this information would be likely to deduce erroneous implications (such as delay propagation through blocks, which is not accurate).

- **Internal Stops**: Many rail operators have trains make stops at certain stations for employees only, be it at a train yard or regular passenger station. Some of these are advertised publicly and announced as an employee-only stop (e.g. "Hillside Support Facility for Employees" on the LIRR), and others aren't advertised whatsoever (e.g. our trains at several different train yards like the Boston Engine Terminal, NJ Transit trains at their Meadowlands Maintenance Complex, SEPTA trains at several train yards, etc.). While they are included in our internal schedule(s), we and most other rail operators do not wish to include them publicly… yet they can't be entered in deadheads.txt as a deadhead trip since this is an internal employee stop on an existing _public_ trip.

- **Waypoints**: Beyond the above, many rail carriers operate trips with times included at intermediate waypoints (e.g. interlockings/switches), but that are not released publicly for security/sensitivity reasons. The same is also seen on light rail, rapid transit, and even some bus systems, albeit less frequently.

> I recognize that some operations already simply release this information publicly in the blocks field and using `pickup_type,drop_off_type` values of `1,1` — albeit with no distinction as to whether that's a _stop_ or merely a passing time — but this public release is simply not viable for us, nor many other operations… particularly on the rail side.

Likewise, there may be scenarios where public information differs from internal information, such as:

- **Internal Routes**: While rail trips on certain corridors/segments may be cobranded to the public as a single line (e.g. our "Franklin/Foxboro Line"), the line may be split into multiple subsets or even to other existing routes for internal purposes (e.g. "Franklin via NEC," "Franklin via DB," "Foxboro via NEC," "Foxboro via DB," "Football Extra")

[...]

PR #61, opened 2024-02-28
Restructure ODS to "Supplement" Framework

jeffkessler-keolis:
Proposed documentation for implementation per #55 and discussion on prior working group calls.

* Establishes the standard structure for `_supplement` files and enumerates those provisioned for use with standardized fields.
* Updates GTFS reference links to the GTFS repo itself (vs the deprecated Google Transit webpage).
* Updates the reference language in this file to TODS (for Transit Operational Data Standard per 2/27/24 call), although other files remain to be updated.
* Leaves the runs_pieces.txt and run_events.txt files unmodified in anticipation of a pull request to implement the changes of #60.

An open question remains as to the preferred handling of the `TODS_trip_type` and `TODS_location_type` fields, both in their nomenclature and their implementation as a text string (vs enum). I have proposed the two above based on handling in the analogous GTFS files and prior working group discussion about the value of text strings vs enums for varying types needing to be defined by individual operators, although these could certainly be changed per future discussions.

### DIFF
--- a/docs/spec/index.md
+++ b/docs/spec/index.md
@@ -1,54 +1,63 @@
+
 # Reference
 
-The Operational Data Standard was last updated on April 14, 2022 (v1.0). View the full [revision history](./revision-history.md).
+The Transit Operational Data Standard was last updated on February 27, 2024 (v2.0). View the full [revision history](./revision-history.md).
+
 
 ## Dataset Files
 
-| **File Name** | **Description** |
-| --- | --- |
-| deadheads.txt | Defines scheduled deadheads contained in a transit feed. (This file is analogous to [trips.txt](https://developers.google.com/transit/gtfs/reference#tripstxt) for non-revenue operations.) |
-| ops_locations.txt | Significant operational locations relevant to the performance of vehicle deadheads. (This file is analogous to [stops.txt](https://developers.google.com/transit/gtfs/reference#stopstxt) for non-revenue operations.) |
-| deadhead_times.txt | Times that a vehicle arrives at and departs from operational locations for each deadhead. (This file is analogous to [stop_times.txt](https://developers.google.com/transit/gtfs/reference#stop_timestxt) for non-revenue operations.) |
-| runs_pieces.txt | Defines daily personnel schedules within a feed. |
-| run_events.txt | Defines other scheduled activities to be performed by a member of personnel during a run. |
+### Structure
 
-## Field Definitions
+There are two types of files used in the TODS standard:
 
-### deadheads.txt
+- **`_supplement` files**, used to add, modify, and delete information from public GTFS files to model the operational service for internal purposes.
+- **TODS-specific files**, used to model operational elements not currently defined in the GTFS standard.
 
-| **Field Name** | **Type** | **Required** | **Description** |
-| --- | --- | --- | --- |
-| deadhead_id | ID | Required | Identifies a deadhead. |
-| service_id | ID referencing [**calendar.service_id**](https://developers.google.com/transit/gtfs/reference#calendartxt) | Required | Identifies a set of dates when the deadhead is scheduled to take place. |
-| block_id | ID | Required | Identifies the block to which the deadhead belongs. |
-| shape_id | ID referencing [**shapes.shape_id**](https://developers.google.com/transit/gtfs/reference#shapestxt) | Optional | Identifies a geospatial shape that describes the vehicle travel path for a deadhead. |
-| to_trip_id | ID referencing [**trips.trip_id**](https://developers.google.com/transit/gtfs/reference#tripstxt) | Conditionally Required | Identifies the trip scheduled immediately following to the deadhead within the block_id. |
-| from_trip_id | ID referencing [**trips.trip_id**](https://developers.google.com/transit/gtfs/reference#tripstxt) | Conditionally Required | Identifies the trip scheduled immediately prior to the deadhead within the block_id. |
-| to_deadhead_id | ID referencing **deadheads.deadhead_id** | Conditionally Required | Identifies the deadhead scheduled immediately following the deadhead within the block_id. |
-| from_deadhead_id | ID referencing **deadheads.deadhead_id** | Conditionally Required | Identifies the deadhead scheduled immediately prior to the deadhead within the block_id. |
 
-### ops_locations.txt
+### Files
+  
+| **File Name** | **Type** | **Description** |
+| --- | --- | --- |
+| trips_supplement.txt | Supplement | Supplements and modifies GTFS [trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) with deadheads and other non-public trip information.|
+| stops_supplement.txt | Supplement | Supplements and modifies GTFS [stops.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) with internal stop locations, waypoints, and other non-public stop information.|
+| stop_times_supplement.txt | Supplement | Supplements and modifies GTFS [stop_times.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) with non-public times that trips stop at locations, and related information. |
+| routes.txt | Supplement | Supplements and modifies GTFS [routes.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#routestxt) with internal route identifiers and other non-public route identification. |
+| runs_pieces.txt | TODS-Specific | Defines daily personnel schedules within a feed. |
+| run_events.txt | TODS-Specific| Defines other scheduled activities to be performed by a member of personnel during a run. |
 
-| **Field Name** | **Type** | **Required** | **Description** |
-| --- | --- | --- | --- |
-| ops_location_id | ID | Required | Identifies an operational location. |
-| ops_location_code | String | Optional | Short text or a number that identifies the operational location for internal use. |
-| ops_location_name | String | Required | Name of the operational location. |
-| ops_location_desc | String | Optional | Description of the operational location. |
-| ops_location_lat | Latitude | Required | The latitude of the operational location. |
-| ops_location_lon | Longitude | Required | The longitude of the operational location. |
+The use of the `_supplement` standard to modify other GTFS files remains experimental at this time.
 
-### deadhead_times.txt
 
-| **Field Name** | **Type** | **Required** | **Description** |
-| --- | --- | --- | --- |
-| deadhead_id | ID referencing **deadheads.deadhead_id** | Required | Identifies a deadhead. |
-| arrival_time | Time | Required | The time at which the vehicle is scheduled to arrive at the operational location specified by this record. |
-| departure_time | Time | Required | The time at which the vehicle is scheduled to depart from the operational location specified by this record. |
-| ops_location_id | ID referencing **ops_locations.ops_location_id** | Conditionally Required | Identifies the operational location specified by this record.<br /><br />If stop_id is blank, ops_location_id is required. If stop_id is not blank, ops_location_id must be blank. |
-| stop_id | ID referencing **[stops.stop_id](https://developers.google.com/transit/gtfs/reference#stopstxt)** | Conditionally Required | Identifies the stop specified by this record.<br /><br />If ops_location_id is blank, stop_id is required. If ops_location_id is not blank, stop_id must be blank. |
-| location_sequence | Non-negative Integer | Required | Order of locations, including both operational locations and stops, for a particular deadhead. The values must increase along the trip but do not need to be consecutive. |
-| shape_dist_traveled | Non-negative Float | Optional | Actual distance traveled along the associated shape, from the first location to the location specified in this record. |
+## `_supplement` Files
+
+### Structure & Evaluation
+
+The fields in all `_supplement` files match those defined in the corresponding file's [GTFS specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md).
+
+Each row in a `_supplement` file is used to modify the corresponding file in GTFS and should be evaluated as follows:
+
+1. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is equal to `1`, remove the row from the GTFS file.
+2. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is not defined or not equal to `1`, set or update (in the corresponding GTFS file) the value of any non-blank fields defined in the `_supplement` file's row to the `_supplement`'s value.
+3. If the row's Primary key is NOT defined in the corresponding GTFS file, add the row to the corresponding GTFS file.
+
+In other words, where a primary key matches, the row is either removed or any non-empty values in the row are used to *update* the corresponding GTFS values. Where a primary key match does not exist, the entire row is added.
+
+### TODS-Specific Fields
+
+In addition to the fields defined in GTFS, specific fields for use within TODS are denoted by a `TODS_` field prefix.
+
+| **File** | **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- | --- |
+| Any `_supplement` file | `TODS_delete` | Enum | Optional | (blank) - Update other fields; do not delete.<br>`1` - Deletes the GTFS row in the corresponding file whose primary key matches the value in the `_supplement` file's row. |
+| `trips_supplement.txt` | `TODS_trip_type` | Text | Optional | Defines the type of the trip if distinct from a standard revenue trip. |
+| `stops_supplement.txt` | `TODS_location_type` | Text | Optional | Defines the type of the location if distinct from a standard GTFS location type. Where defined, the GTFS `location_type` shall be ignored. |
+
+
+
+
+
+## TODS-Specific File Definitions
+
 
 ### runs_pieces.txt
 
@@ -57,11 +66,11 @@ The Operational Data Standard was last updated on April 14, 2022 (v1.0). View th
 | run_id | ID | Required | Identifies a run. |
 | piece_id | ID | Required | Identifies the piece of the run. The piece_id field must be unique. |
 | start_type | Enum | Required | Indicates whether the piece begins with a deadhead, a revenue trip, or an event.<br /><br />**0** - Deadhead <br />**1** - Trip <br />**2** - Event |
-| start_trip_id | ID referencing **deadheads.deadhead_id** or [**trips.trip_id**](https://developers.google.com/transit/gtfs/reference#tripstxt) | Required | Identifies the deadhead or trip with which the piece begins. |
-| start_trip_position | Non-negative Integer referencing **deadhead_times.location_sequence** or [**stop_times.stop_sequence**](https://developers.google.com/transit/gtfs/reference#stop_timestxt) | Optional | Identifies the first operational location or stop to be serviced in the first trip of the piece. This field should only be filled out if the piece does not begin at the first stop of the start trip. |
+| start_trip_id | ID referencing **deadheads.deadhead_id** or [**trips.trip_id**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) | Required | Identifies the deadhead or trip with which the piece begins. |
+| start_trip_position | Non-negative Integer referencing **deadhead_times.location_sequence** or [**stop_times.stop_sequence**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) | Optional | Identifies the first operational location or stop to be serviced in the first trip of the piece. This field should only be filled out if the piece does not begin at the first stop of the start trip. |
 | end_type | Enum | Required | Indicates whether the piece ends with a deadhead, a revenue trip, or an event. <br /><br />**0** - Deadhead <br />**1** - Trip <br />**2** - Event |
-| end_trip_id | ID referencing **deadheads.deadhead_id** or [**trips.trip_id**](https://developers.google.com/transit/gtfs/reference#tripstxt) | Required | Identifies the deadhead or trip with which the piece ends. |
-| end_trip_position | Non-negative Integer referencing **deadhead_times.location_sequence** or [**stop_times.stop_sequence**](https://developers.google.com/transit/gtfs/reference#stop_timestxt) | Optional | Identifies the last operational location or stop to be serviced in the last trip of the piece. This field should only be filled out if the piece does not end at the last stop of the end trip. |
+| end_trip_id | ID referencing **deadheads.deadhead_id** or [**trips.trip_id**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) | Required | Identifies the deadhead or trip with which the piece ends. |
+| end_trip_position | Non-negative Integer referencing **deadhead_times.location_sequence** or [**stop_times.stop_sequence**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) | Optional | Identifies the last operational location or stop to be serviced in the last trip of the piece. This field should only be filled out if the piece does not end at the last stop of the end trip. |
 
 ### run_events.txt
 
@@ -74,6 +83,6 @@ The Operational Data Standard was last updated on April 14, 2022 (v1.0). View th
 | event_time | Time | Required | The time at which the event begins. |
 | event_duration | Non-negative Integer | Required | The scheduled duration of the event from the event_time in seconds. |
 | event_from_location_type | Enum | Optional | Indicates whether the event is scheduled to begin at an operational location or a stop.<br /><br />**0** - Operational Location<br />**1** - Stop |
-| event_from_location_id | ID referencing **ops_locations.ops_location_id** or [**stops.stop_id**](https://developers.google.com/transit/gtfs/reference#stopstxt) | Optional | Identifies the operational location or stop at which the event is scheduled to begin. |
+| event_from_location_id | ID referencing **ops_locations.ops_location_id** or [**stops.stop_id**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) | Optional | Identifies the operational location or stop at which the event is scheduled to begin. |
 | event_to_location_type | Enum | Optional | Indicates whether the event is scheduled to end at an operational location or a stop.<br /><br />**0** - Operational Location<br />**1** - Stop |
-| event_to_location_id | ID referencing **ops_locations.ops_location_id** or [**stops.stop_id**](https://developers.google.com/transit/gtfs/reference#stopstxt) | Optional | Identifies the operational location or stop at which the event is scheduled to end. |
+| event_to_location_id | ID referencing **ops_locations.ops_location_id** or [**stops.stop_id**](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) | Optional | Identifies the operational location or stop at which the event is scheduled to end. |# Reference
--- a/docs/spec/index.md
+++ b/docs/spec/index.md
@@ -42,6 +42,10 @@ Each row in a `_supplement` file is used to modify the corresponding file in GTF
 
 In other words, where a primary key matches, the row is either removed or any non-empty values in the row are used to *update* the corresponding GTFS values. Where a primary key match does not exist, the entire row is added.
 
+> _**Implications and Guidance:**_
+> - To explicitely blank a field, delete the row entirely using the `TODS_delete` field and then re-add the row with the desired values subsequent thereto.
+> - If a row contains defined values besides the Primary Key and a `TODS_delete` value of `1`, the row shall be removed and other values in that row will be ignored.
+
 ### TODS-Specific Fields
 
 In addition to the fields defined in GTFS, specific fields for use within TODS are denoted by a `TODS_` field prefix.


docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+
+1. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is equal to `1`, remove the row from the GTFS file.
+2. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is not defined or not equal to `1`, set or update (in the corresponding GTFS file) the value of any non-blank fields defined in the `_supplement` file's row to the `_supplement`'s value.
+3. If the row's Primary key is NOT defined in the corresponding GTFS file, add the row to the corresponding GTFS file.
+
+In other words, where a primary key matches, the row is either removed or any non-empty values in the row are used to *update* the corresponding GTFS values. Where a primary key match does not exist, the entire row is added.
+
+> _**Implications and Guidance:**_
+> - To explicitely blank a field, delete the row entirely using the `TODS_delete` field and then re-add the row with the desired values subsequent thereto.
+> - If a row contains defined values besides the Primary Key and a `TODS_delete` value of `1`, the row shall be removed and other values in that row will be ignored.
```
skyqrose, line 47: This all seems correct, but I had a really hard time reading it. Part of that is cuz the rules are kind of complicated, but maybe a rewrite could make it clearer?

Some more specific suggestions:
- The main idea of adding+updating+deleting rows gets obscured by all the details. Perhaps there should be a simple explanation first that glosses over some details, followed by a less-readable full list of rules?
- The behavior of blank fields is barely mentioned, but will be a very important thing that people use all the time. Perhaps it deserves more weight, especially in the readable-but-imprecise guide.
- "Primary Key" may require a bit more explanation. It's not something that matters too much for most people's use of GTFS, so people might not be familiar with it. Maybe just a link to [the GTFS docs about it](https://gtfs.org/schedule/reference/#dataset-attributes), or a "The columns that make up a file's Primary Key are listed with each file in its docs" explanation or something.
- Typo: "explicitely" -> "explicitly".

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+Each row in a `_supplement` file is used to modify the corresponding file in GTFS and should be evaluated as follows:
+
+1. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is equal to `1`, remove the row from the GTFS file.
+2. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is not defined or not equal to `1`, set or update (in the corresponding GTFS file) the value of any non-blank fields defined in the `_supplement` file's row to the `_supplement`'s value.
+3. If the row's Primary key is NOT defined in the corresponding GTFS file, add the row to the corresponding GTFS file.
+
+In other words, where a primary key matches, the row is either removed or any non-empty values in the row are used to *update* the corresponding GTFS values. Where a primary key match does not exist, the entire row is added.
+
+> _**Implications and Guidance:**_
+> - To explicitely blank a field, delete the row entirely using the `TODS_delete` field and then re-add the row with the desired values subsequent thereto.
```
skyqrose, line 46: If the only way to remove a single column is to delete the row and recreate it, that seems like an oversight in the spec. Can we fix it?

Some specific reasons that's bad:
- It's clunky to write two lines instead of one. Especially if you need to delete an entire column for all rows. The file would end up being twice as long as the GTFS it amends. 
- The primary key would then show up twice in this file. If the order is wrong, then you could delete the row you just modified. GTFS files should have unique primary keys and be order-independent.
- Requires duplicating any columns that you want to save unmodified from GTFS, instead of leaving the field blank.

Two ideas for a solution:
1. Any GTFS column could be given a value of "ODS_blank" (and hope nobody every has a stop with exactly that name)
2. Instead of `ODS_delete` being a simple boolean, it's a list of columns to delete

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+
+## `_supplement` Files
+
+### Structure & Evaluation
+
+The fields in all `_supplement` files match those defined in the corresponding file's [GTFS specification](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md).
+
+Each row in a `_supplement` file is used to modify the corresponding file in GTFS and should be evaluated as follows:
+
+1. If the row's Primary Key is defined in the corresponding GTFS file, and the `TODS_delete` field is equal to `1`, remove the row from the GTFS file.
```
skyqrose, line 39: As came up in the meeting on Tuesday, there should be some explanation of what happens if you delete a trip (or other files referenced by foreign keys). Should the stop_times be deleted, too, or just left hanging without a trip and ignored?

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+
+### TODS-Specific Fields
+
+In addition to the fields defined in GTFS, specific fields for use within TODS are denoted by a `TODS_` field prefix.
+
+| **File** | **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- | --- |
+| Any `_supplement` file | `TODS_delete` | Enum | Optional | (blank) - Update other fields; do not delete.<br>`1` - Deletes the GTFS row in the corresponding file whose primary key matches the value in the `_supplement` file's row. |
+| `trips_supplement.txt` | `TODS_trip_type` | Text | Optional | Defines the type of the trip if distinct from a standard revenue trip. |
+| `stops_supplement.txt` | `TODS_location_type` | Text | Optional | Defines the type of the location if distinct from a standard GTFS location type. Where defined, the GTFS `location_type` shall be ignored. |
```
skyqrose, line 57: Then why not modify the `stops_supplement.txt/location_type`?

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+> - If a row contains defined values besides the Primary Key and a `TODS_delete` value of `1`, the row shall be removed and other values in that row will be ignored.
+
+### TODS-Specific Fields
+
+In addition to the fields defined in GTFS, specific fields for use within TODS are denoted by a `TODS_` field prefix.
+
+| **File** | **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- | --- |
+| Any `_supplement` file | `TODS_delete` | Enum | Optional | (blank) - Update other fields; do not delete.<br>`1` - Deletes the GTFS row in the corresponding file whose primary key matches the value in the `_supplement` file's row. |
+| `trips_supplement.txt` | `TODS_trip_type` | Text | Optional | Defines the type of the trip if distinct from a standard revenue trip. |
```
skyqrose, line 56: What values would go here? Is this for deadheads? It might be nice to have a `is_in_passenger_service` boolean field that doesn't require matching against arbitrary text. That could replace this or be in addition to it.

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
-| ops_location_lat | Latitude | Required | The latitude of the operational location. |
-| ops_location_lon | Longitude | Required | The longitude of the operational location. |
+| **File Name** | **Type** | **Description** |
+| --- | --- | --- |
+| trips_supplement.txt | Supplement | Supplements and modifies GTFS [trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) with deadheads and other non-public trip information.|
+| stops_supplement.txt | Supplement | Supplements and modifies GTFS [stops.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) with internal stop locations, waypoints, and other non-public stop information.|
+| stop_times_supplement.txt | Supplement | Supplements and modifies GTFS [stop_times.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) with non-public times that trips stop at locations, and related information. |
+| routes.txt | Supplement | Supplements and modifies GTFS [routes.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#routestxt) with internal route identifiers and other non-public route identification. |
```
safrazier17, line 24: This should be `routes_supplement.txt`

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+| runs_pieces.txt | TODS-Specific | Defines daily personnel schedules within a feed. |
+| run_events.txt | TODS-Specific| Defines other scheduled activities to be performed by a member of personnel during a run. |
 
-### deadhead_times.txt
+The use of the `_supplement` standard to modify other GTFS files remains experimental at this time.
+
+
+## `_supplement` Files
```
safrazier17, line 31: I think we should keep the practice of having individual tables for each file, even if the `_supplement` files have the same information for now.

skyqrose: As long as the tables don't have to list every GTFS column, and can elide them. Listing out all the columns in all the files and fields would be so much text that it'd make it hard to see where the new and interesting parts of TODS are, and would make it harder to understand the shared ideas behind the supplement files.

Example:

`trips_supplement.txt`

<table>
<tr>
<th>Field name</th>
<th>Type</th>
<th>Required</th>
<th>Description</th>
</tr>
<tr>
<td>... any column in GTFS `trips.txt`</td>
<td>same values as in GTFS</td>
<td>Conditionally Required</td>
<td>
Required if adding new data to GTFS, and the GTFS column is required.
Optional if the GTFS field is optional, or if it's modifying a required row that already exists GTFS.
Should be blank if `TODS_delete` is `1`.
</td>
</tr>
<tr>
<td><code>TODS_trip_type</code></td>
<td>Text</td>
<td>Optional</td>
<td>
Defines the type of the trip if distinct from a standard revenue trip.
</td>
</tr>
</tr>
<tr>
<td><code>TODS_delete</code></td>
<td>Enum</td>
<td>Optional</td>
<td>
If <code>1</code>, removes the row from GTFS instead of modifying it.
</td>
</tr>
</table>

Although I think that even having 4 near-identical tables for the supplement files might make it hard to spot the differences.

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+
+| **File** | **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- | --- |
+| Any `_supplement` file | `TODS_delete` | Enum | Optional | (blank) - Update other fields; do not delete.<br>`1` - Deletes the GTFS row in the corresponding file whose primary key matches the value in the `_supplement` file's row. |
+| `trips_supplement.txt` | `TODS_trip_type` | Text | Optional | Defines the type of the trip if distinct from a standard revenue trip. |
+| `stops_supplement.txt` | `TODS_location_type` | Text | Optional | Defines the type of the location if distinct from a standard GTFS location type. Where defined, the GTFS `location_type` shall be ignored. |
+
+
+
+
```
skyqrose, line 61: I think a small example would help show how the supplement idea works. I know that examples are usually separate from the reference, but I think this idea would be significantly easier to explain with an example, so it's worth adding one here.

### Example

GTFS `stops.txt`:
```csv
stop_id,stop_name,stop_desc,stop_url
1,One,Unmodified in TODS,example.com/1
2,Two,Deleted in TODS,example.com/2
3,Three,Will be modified in TODS,example.com/3
```

TODS `stops_supplement.txt`
```csv
stop_id,stop_name,stop_desc,TODS_delete
2,,,1
3,,Has been modified by TODS,
4,Four,New in TODS,
```

Effective `stops.txt` after merging in the supplement file:
```csv
stop_id,stop_name,stop_desc,stop_url
1,One,Unmodified in TODS,example.com/1
3,Three,Has been modified by TODS,example.com/3
4,Four,New in TODS,
```

Note that the station name "Three" was not modified, and the whole column `stop_url` was omitted and not modified.

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
-| ops_location_name | String | Required | Name of the operational location. |
-| ops_location_desc | String | Optional | Description of the operational location. |
-| ops_location_lat | Latitude | Required | The latitude of the operational location. |
-| ops_location_lon | Longitude | Required | The longitude of the operational location. |
+| **File Name** | **Type** | **Description** |
+| --- | --- | --- |
+| trips_supplement.txt | Supplement | Supplements and modifies GTFS [trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) with deadheads and other non-public trip information.|
+| stops_supplement.txt | Supplement | Supplements and modifies GTFS [stops.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) with internal stop locations, waypoints, and other non-public stop information.|
+| stop_times_supplement.txt | Supplement | Supplements and modifies GTFS [stop_times.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) with non-public times that trips stop at locations, and related information. |
```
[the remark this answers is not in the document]
jeffkessler-keolis: How about "with non-public times **_at which_** trips stop at locations, and related information" (emphasis added)?

The point is that it doesn't include the locations, but rather the non-public times that they're stopping at these locations.

docs/spec/index.md
jeffkessler-keolis: Very good catch!

docs/spec/index.md
jeffkessler-keolis: My thought process behind this was to think from the perspective of someone directly interacting with the data.

By presenting it as, "all files of this type do the same thing from x to y," it guides a reader to implementing a common programatic framework for the parsing and processing of each file rather than a bespoke process specific to each file.

This was also the motivation behind the note about the experimental application to other files: basically hinting at the notion that subsequent file additions of this type will follow the same approach/methodology, so implementing a generalized approach will save effort in the long-term.

My concerns with listing each individual file are:
1. That's a lot of data to either replicate from GTFS or include that's redundant.
2. Similarities between files would be more difficult to spot (per Sky).
3. It could prompt developers to implement different processes for different files, which could present challenges in the future.
4. As additional files are added, redundancies would grow and grow, presenting additional opportunities for error.

docs/spec/index.md
jeffkessler-keolis: This provides us the flexibility to implement the industry-specific location type jargon needed for our individual operations and classifications without breaking the existing GTFS data structures.

If we were to modify it directly, the values being modified in `stops.txt/location_type` would change from an enum to a string and would not be accepted by a validator. This keeps the TODS-specific datatype changes in non-GTFS fields.

docs/spec/index.md
jeffkessler-keolis: I personally don't see a use case for blanking a field, since it's not the case where an operator might want to obscure internal data — this is a spec for adding non-public information to public information; anything extraneous need not be removed from the existing feed, but could simply ignore it in an internal consumer application.

The concept of replicating an entire row seems like a non-starter to me. At that point, we're basically back to publishing internal and external GTFS feeds, and many — myself included — would simply just follow that practice if I already needed to worry about validating each specific field in two files.

Does anyone have a specific use case for why they'd need to delete a column value (without replacing it)? Otherwise, I'm inclined to say we should ignore for now.

jeffkessler-keolis: (The only reason I added this note is from a discussion with an industry colleague doing a once-over who made the observation. I do like the premise of having things processed in-order to ensure things can be properly replicated. Or to preserve order-independence, per Sky's comment, should we explicitly prohibit this behavior and say a primary key that's already used in GTFS can't be deleted and re-added via TODS?)

docs/spec/index.md
```diff
@@ -1,54 +1,67 @@
...
+| routes.txt | Supplement | Supplements and modifies GTFS [routes.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#routestxt) with internal route identifiers and other non-public route identification. |
+| runs_pieces.txt | TODS-Specific | Defines daily personnel schedules within a feed. |
+| run_events.txt | TODS-Specific| Defines other scheduled activities to be performed by a member of personnel during a run. |
 
-### deadhead_times.txt
+The use of the `_supplement` standard to modify other GTFS files remains experimental at this time.
```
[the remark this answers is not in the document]
jeffkessler-keolis: I would consider it akin to the GTFS definition of:

> **Experimental fields**
> Fields labeled as experimental are subject to change and not yet formally adopted into the specification. An experimental field may be formally adopted in the future.

Perhaps this should be adopted into the formal governance (or listed herein)?

docs/spec/index.md
jeffkessler-keolis: Good points!
- I split "Structure" into a new section to include more of the core idea before getting into the evaluation process.
- Blank field detail added to "Implications & Guidance"
- Primary Key described and linked as suggested
- Typo corrected

docs/spec/index.md
jeffkessler-keolis: Added to Implications & Guidance; feel free to comment therein if you feel strongly that all corresponding rows should be removed.

docs/spec/index.md
jeffkessler-keolis: Great example; added!

docs/spec/index.md
jeffkessler-keolis: I like that concept! Thoughts on that terminology or something shorter like `TODS_revenue`?

I realize some operations may be fare-free, but the term "revenue service" is fairly well-defined in the US market, particularly given the FTA definition:

> Revenue Service (Miles, Hours, and Trips)
> The time when a vehicle is available to the general public and there is an expectation of carrying passengers. These passengers either: 
> •   Directly pay fares;
> •   Are subsidized by public policy; or
> •   Provide payment through some contractual arrangement.
>  
> Vehicles operated in fare free service are considered in revenue service. Revenue service includes: 
> •   Layover / recovery time.
>  
> Revenue service excludes: 
> •   Deadhead;
> •   Vehicle maintenance testing;
> •   School bus service; and
> •   Charter service.
> Can be found in: A-10, A-30, S-10, R-20, FFA-10, Declarations

docs/spec/index.md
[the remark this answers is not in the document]
jeffkessler-keolis: Detail added in the latest version

docs/spec/index.md
skyqrose: I think Primary Keys should be unique in these files, and we shouldn't allow a duplicate row even if one is to delete.

I don't have a specific use case for blanking fields, but my hypothetical use case isn't about obscuring data, but about removing inaccurate data. Hypothetically, maybe an agency schedules bus stop times only at timepoints, and adds interpolated times at other stops to the public GTFS, but doesn't want to have the interpolated times in the internal feed.

I'm fine with not allowing blanking a field now, as long as we think we can solve it in a backwards-compatible way if it does come up later.

I looked through the GTFS docs, and didn't see any places we'd end up with an invalid GTFS file if we couldn't blank fields, even if we expand to other GTFS files. All optional enums have a value that's equivalent to blank, and all conditionally forbidden fields wouldn't have a way to become forbidden by adding TODS files.

docs/spec/index.md
skyqrose: Makes sense. Is there a way to include that answer in the docs somewhere?

docs/spec/index.md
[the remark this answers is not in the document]
skyqrose: I think Josh's point was that this file also includes new stop_times for new non-public locations, not just modifications to stop_times at public locations. The docs could suggest that use case. (I see how you're trying to avoid making it sound like this file contains the data about those locations, cuz that's in stops_supplement, but I don't think that'll be confusing.)

docs/spec/index.md
jfabi: ➕ Agreed with @skyqrose on further review, it's probably not worth trying to jam a solution right now; the other strongest example I could come up with is changing a route designation internally, and wanting to nullify a `route_long_name` or `route_short_name`, but it might be more likely you'd just move to use a new `route_id` anyway where you could dictate that.

docs/spec/index.md
jeffkessler-keolis: As I've tried modifying things for this, I am not sure if we should include this. Some vendors may which to use this boolean methodology to distinguish revenue trips from nonrevenue trips, but I think the better approach is letting each agency specify which values should be considered as revenue or nonrevenue service.

Perhaps that becomes a future TODS-specific file of `trip_types.txt` that maps specific values of the `TODS_trip_type` field to a revenue/nonrevenue or other related categorization that we can open as a new problem.

docs/spec/index.md
jeffkessler-keolis: In reading through things, that seems to belong more in guidance not for _use_ of the spec, but in governance of the specification. Perhaps a new eponymous "Guidance" section of governance.txt is appropriate here?

docs/spec/index.md
```diff
@@ -1,54 +1,104 @@
...
+### Files
 
-### ops_locations.txt
+| **File Name** | **Type** | **Description** |
+| --- | --- | --- |
+| trips_supplement.txt | Supplement | Supplements and modifies GTFS [trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) with deadheads and other non-public trip information. |
+| stops_supplement.txt | Supplement | Supplements and modifies GTFS [stops.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) with internal stop locations, waypoints, and other non-public stop information.|
+| stop_times_supplement.txt | Supplement | Supplements and modifies GTFS [stop_times.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) with non-public times at which trips stop at locations, and related information. |
+| routes_supplement.txt | Supplement | Supplements and modifies GTFS [routes.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#routestxt) with internal route identifiers and other non-public route identification. |
```
skyqrose, line 22: I'm working on updating the examples to the new spec, and something I ran into: What route are deadheads on?

It's a little awkward cuz routes don't describe a path a vehicle takes. Instead, they describe a public-facing name that a trip has. But deadheads aren't public facing, so they don't have an identity like that.

Should the deadhead's route be the route that it's pulling out to? Some global "deadhead" route? A different route for every different type of deadhead? Can we make the `trips_supplement.txt/route_id` field optional?

jeffkessler-keolis: I don't think this is something that we should define, as most existing agencies already have a standard practice per their scheduling system that they can continue to use.

For example, some systems (a) assign the trips to the route of their adjacent revenue trip, some (b) assign it to a "deadhead" route, some (c) assign it to a specific deadhead route by garage; all of those approaches are supported in the specification provided the routes are defined in routes_supplement.

Likewise, I think the more that we stray from the GTFS structure (i.e. saying things in static GTFS are not required), the more opportunity there is for confusion or breaking of existing architecture, so I would not recommend making it optional.

I think a fairly universal principle we should uphold is that the set of Supplemented GTFS files (static GTFS schedule files after having applied the Supplement modifications) should still validate in accordance in accordance with the canonical GTFS validator.

Perhaps it's worth:
- Including the above definition of Supplemented GTFS (Static GTFS after TODS Supplement modifications), and the notion that it should validate, in the documentation?
- Adding a "best practices" section akin to the equivalent in static GTFS, or perhaps even something akin to a "statement of legislative intent" that records these decisions being made with intentionality?

docs/spec/index.md
jfabi: One possible reading of the definition as-written is that if `TODS_trip_type` is null, the supplemental trip should be treated as a standard revenue trip. Is that the intended default, or is it worth clarifying (for example, that there is no default beyond it being non-public information)?

jeffkessler-keolis:
Just bumping this thread since it seems the v2 specification proposal has been stable with no further commentary over the past few months.

@safrazier17 @Cristhian-HA @jfabi Do you know what'd be the next step in progressing the formal adoption of this and @skyqrose's #60? Can we proceed with calling for a vote?

(In terms of the sequencing of things, I think we discussed on the last call to vote on the combined adoption, then merging this in first, followed by Sky's change which includes some references to fields/data herein.)

Thanks!

safrazier17:
### VOTING PERIOD OPEN
Voting is open on this PR through 11:59:59 UTC on June 19, 2024. This PR is the first of 2 comprising the TODS 2.0 target release. If adopted, this PR will be versioned as TODS 2.0-alpha.1.

### WHO IS ELIGIBLE TO VOTE

Issue Working Group members @safrazier17 @jeffkessler-keolis @jfabi @skyqrose are not eligible to vote on this proposal.

### HOW TO VOTE

### OUTCOMES

wmata-eajensen:
+1 WMATA

paulswartz (commented):
Pending one text clarification, +1 MBTA.

docs/spec/index.md
```diff
@@ -1,54 +1,105 @@
...
+1,One,Unmodified in TODS,example.com/1
+3,Three,Has been modified by TODS,example.com/3
+4,Four,New in TODS,
+```
+
+_Note that the station name "Three" was not modified, and the whole column stop_url was omitted and not modified._
+
+### Implications and Guidance
+
+- As blank fields are ignored, data to be removed should either be overwritten with a new value or have their entire row deleted using the `TODS_delete` field.
```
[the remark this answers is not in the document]
jeffkessler-keolis: See [https://github.com/cal-itp/operational-data-standard/pull/61#discussion_r1513587956](https://github.com/cal-itp/operational-data-standard/pull/61#discussion_r1513587956).

This is not currently possible in the proposed version of the spec, but since a consumer could simply ignore this data, we struggled to find a use case that warranted immediate inclusion.

I think @skyqrose's proposal of a `tods_blank` field is the most promising, but has some implementation details worth reviewing before including.

The consensus was to move forward and track this as an open item should it present itself as an item needing to be addressed in a subsequent update.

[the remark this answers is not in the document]
paulswartz: Thanks! That discussion addresses my concerns.

hunterowens:
+1 Cal-ITP / Caltrans!

docs/spec/index.md
antrim: @jeffkessler-keolis @skyqrose 
**RE: TODS_delete:** In what cases is it necessary to remove (rather than supplement) passenger-facing information from the GTFS? This caught my eye out of a desire to keep implementation simple.

docs/spec/index.md
```diff
@@ -1,54 +1,105 @@
...
+### Files
 
-### ops_locations.txt
+| **File Name** | **Type** | **Description** |
+| --- | --- | --- |
+| trips_supplement.txt | Supplement | Supplements and modifies GTFS [trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt) with deadheads and other non-public trip information. |
+| stops_supplement.txt | Supplement | Supplements and modifies GTFS [stops.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt) with internal stop locations, waypoints, and other non-public stop information.|
+| stop_times_supplement.txt | Supplement | Supplements and modifies GTFS [stop_times.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt) with non-public times at which trips stop at locations, `stop_times` entries for non-public trips, and related information. |
+| routes_supplement.txt | Supplement | Supplements and modifies GTFS [routes.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#routestxt) with internal route identifiers and other non-public route identification. |
```
mads14, line 22: Swiftly has a loose preference to make the file suffix be _tods_supplement rather than _supplement, given that there are a couple of optional columns that start with TODS_ in those supplement files. 

We would rather have multiple supplement files per original file than a single supplement file that feels "sparse", i.e. many columns are always null for many rows, because there's a different set of columns targeted at different reasons for supplementing

jeffkessler-keolis:
Since there was some feedback requesting a summary of the motivation behind this change, the summary is:

- Many large/complex operators, especially rail operators, do not have a clear delineation between revenue/nonrevenue trips.
- The current v1 standard is unusable for these operators since there is no current mechanism to update/modify/add information to trips that already exist in GTFS.
- The supplement structure provides a clean way to update/modify/add information to the content already published in GTFS.
- Since the existing TODS files of deadheads.txt, ops_locations.txt, and deadhead_times.txt are effectively adding non-public information to existing GTFS files (of trips.txt, stops.txt, and stop_times.txt, respectively), their content is moved to the more multipurpose supplement files.

**tl;dr The supplement framework allows complex operators (e.g. rail) to model much greater complexity and combine public and non-public information as required using a standardized approach that builds atop existing GTFS.**

For more details and examples of the motivation behind this change, see https://github.com/cal-itp/operational-data-standard/issues/55.

docs/spec/index.md
jeffkessler-keolis: > Swiftly has a loose preference to make the file suffix be _tods_supplement rather than supplement, given that there are a couple of optional columns that start with TODS in those supplement files.

Makes sense! I'll confer with some of the other working group stakeholders and get back to you.

> We would rather have multiple supplement files per original file than a single supplement file that feels "sparse", i.e. many columns are always null for many rows, because there's a different set of columns targeted at different reasons for supplementing

If I'm understanding this comment correctly, the suggestion proposes allowing a TODS dataset to have two distinct instances of a supplement file (e.g. `trips_supplement_1.txt` and `trips_supplement_2.txt`), such that, for example, one could have the addition of full rows, and another could have some updates to values, without needing to leave the unmodified column values of the applicable update rows blank.

I'm not a huge fan of this from two perspectives:
1. Producers would implicitly be asked to split the content into sub-files (at considerable development effort/expense), which may or may not even be possible in a particular production system, if they even do this at all.
2. Consumers would be required to inspect a directory for multiple sub-files, rather than being able to more easily error-check or report based on the presence of absence of a particular file.

docs/spec/index.md
jeffkessler-keolis: We never identified such a use case; see https://github.com/cal-itp/operational-data-standard/pull/61#discussion_r1513587956 for the thread of that discussion.

docs/spec/index.md
safrazier17: I have no objection to adopting the _tods_supplement name scheme. 

@mads14 could you clarify what the multiple supplement files would look like and how that would help swiftly?

mads14: Thanks for the responses!

To clarify - my suggestion at this point is to rename the files to include _tods_supplement at the end e.g. trips_tods_supplement.txt. My suggestion is not to split up the current proposed supplemental trips file into multiple files.

I imagine that in the future there was a non-TODS extension of GTFS that also requires supplemental trips. 

For example only, I’m not proposing this at the moment: say in the future GTFS was additionally extended so that a substitute “snow schedule” could be placed in supplemental GTFS files and could be swapped in for regular service as needed. This use case would potentially require a different set of additional columns and the TODS_ specific fields would be meaningless/always left blank. Over time we could end up with lots of columns that are only valid for one specific case. I guess the implementers of the next GTFS extension could have a choice and will propose what they will (whether to combine "tods" and the "special schedule" supplements or not) if and when the spec is further extended, but my preferences is that the supplemental files for this new “snow schedule”/“special schedule” are new supplemental files and use their own extension.

safrazier17: Got it, seems like a good change to me @mads14 !

safrazier17:
### VOTING PERIOD CLOSED
The voting period for this proposal closed yesterday, June 19 at 11:59:59 UTC. The proposal received the following votes:

Support (3): @wmata-eajensen (WMATA), @paulswartz (MBTA), @hunterowens (Caltrans)

The Issue Working Group (@jeffkessler-keolis, @skyqrose, @safrazier17, @jfabi) is considering a [minor recommended change](https://github.com/cal-itp/operational-data-standard/pull/61#discussion_r1645182219) from @mads14 (Swiftly) that was not tied to a vote. If the issue working group decides to adopt this change, those voters who cast supportive votes will have a period of four (4) days in which to review and withdraw their support _based on this change_ if they so choose. **This is not a new vote window and no new votes will affect the outcome of this proposal.** 

If withdrawals of support cause the proposal to fall beneath the required threshold of support for passage, the Issue Working Group will decide whether to revert to the previously passed version of the proposal, initiate a new voting window for the proposal, or appeal to the Board for a determination.

If support for the proposal remains above the threshold of support needed for passage, the proposal will immediately be adopted at the end of four days as TODS 2.0-alpha.1.

paulswartz:
I saw @mads14 's proposal and that would not change my vote.

safrazier17:
### FINAL VOTE AND ACTIONS

After further discussion, the Issue Working Group decided **not to adopt the proposed change** of filenames from `_supplement` to `_TODS_supplement` for the following reasons:

1. The Issue Working Group had doubts about whether there were other divisions beyond the internal-facing/external-facing division that would be appropriate for supplementing GTFS without being part of TODS
2. The Issue Working Group had concerns about whether there were logical complications introduced by having multiple files supplementing a single GTFS feed, which had not been fully worked through

As previously announced, the proposal passes with 3 votes in support and none in opposition. The vote and this discussion will be recorded on the website, and TODS 2.0-alpha.1 will be added to the revision history for TODS on the website.

In the coming days, we expect to take up the second of two PRs for the final TODS 2.0 release. That PR is #66 for those interested in reviewing it / making comments before voting opens.

PR #62, opened 2024-02-28
Fix: correct local devcontainer and deployment build

thekaveman:
* Use a non-root user in the devcontainer
* Pin some key dependencies of `mkdocs-material` for deployment builds


e-lo (approved):
I'm not sure WHY it needs all this devcontainer stuff to run the GH actions but let's give it a whirl...

PR #79, opened 2024-08-19
Update revision-history.md to add v2.0.0

safrazier17:


timon-k:
@safrazier17 Looks like this needs a rebase, but it's still worth merging as https://ods.calitp.org/spec/revision-history still does not contain `2.0.0` up to now.

safrazier17:
@timon-k it looks like this will be taken care of in #93

Issue #45, opened 2023-11-30
add rosters.txt


BTollison:
In an effort to help distribute information between platforms, we wish to add rosters.txt. The intended use is that many operations cluster several duties (runs) within a roster to assign to a driver. The duties are usually clustered in a way that represents the amount of work a single driver will work in a week period. Some companies also cluster this work in several weeks for 1 drivers.

**ods_roster_id**
Persistent ID as proposed in issue #44 meant to help keep track of changes to a roster over time. Rosters can have long lifespans while the duties (runs) can change within them over time.
- Required

**roster_id**
- Required, primary key

**service_id**
- Required, defined service period that the duties and roster_id is valid for.

**monday_run**
- Required, value should be a run_id that operates on a monday during this service period.
- If blank, no run_id is assigned.

**tuesday_run**
- Required, value should be a run_id that operates on a tuesday during this service period.
- If blank, no run_id is assigned.

**wednesday_run**
- Required, value should be a run_id that operates on a wednesday during this service period.
- If blank, no run_id is assigned.

**thursday_run**
- Required, value should be a run_id that operates on a thursdsay during this service period.
- If blank, no run_id is assigned.

**friday_run**
- Required, value should be a run_id that operates on a friday during this service period.
- If blank, no run_id is assigned.

**saturday_run**
- Required, value should be a run_id that operates on a saturday during this service period.
- If blank, no run_id is assigned.

**sunday_run**
- Required, value should be a run_id that operates on a saturday during this service period.
- If blank, no run_id is assigned.

**week_sequence**
- if blank, then the roster represents only 1 week.
- This value represents what what week this row of the roster should represent. If a roster_id has 2 weeks worth of work, then there should be a value 1 in the first week and a value 2 in the second week.

skyqrose:
I can see how a grouping of many days' work is useful to represent even when there's no specific person assigned to it yet, as part of the scheduling process. So far, ODS has been mostly focused on data useful during operation, as opposed to during planning+scheduling. Would this be useful during operation (like if you need to check if the usual driver is working today), or is this intended to expand the scope of ODS into scheduling?

Would it be a specific week? I don't see a date field. There's a service_id field, but service_ids can be a many days or only 1, consecutive or not. What does the `sunday_run` field mean if applied to a service_id that only occurs on weekdays? How would you represent how holiday work is grouped with other work?

[...]

Issue #28, opened 2023-11-16
Allocating drivers and vehicles in ODS


hodb:
In reviewing the ODS spec, it appears that there is an absence of guidance regarding the allocation of blocks to operational vehicles, as well as the assignment of duties to drivers with their matching operated dates

As this crucial information is expected to be sourced from the dispatching software, can you provide clarification or additional details on how the ODS spec addresses the extraction of data related to block allocation, driver duties, and associated dates from the dispatching software for communication to the AVL provider?

Thanks

skyqrose:
Matching operators to their duties on each day is something that I would use, and if it's not in ODS, I'd have to represent the data in a non-standard adjacent data feed. I'm currently using a CSV with 3 columns for this purpose: `date`, `run_id`, `operator`.

But it's also kind of separate from the schedules ODS does represent. For example, assigning operators to runs is done well after the rest of the schedule is finalized.

I think I remember this coming up in discussions a while ago and I think it was just cut for scope. It'd be easy to add as a backwards-compatible extension once the rest of the format is finalized.

hodb:
Thanks for your answer @skyqrose. The concept I aim for is pretty much the same - having a new file assignments.txt file that includes also the assignment per drivers and **vehicles**

`block_id,vehicle_id,run_id,driver_id,operational_date`

safrazier17:
> I think I remember this coming up in discussions a while ago and I think it was just cut for scope. It'd be easy to add as a backwards-compatible extension once the rest of the format is finalized.

Yes, this was discussed and cut for scope and also because there seemed to be less consensus on the topic of representing driver/staff-identifying information, at least at that time. I remain in favor of bringing this information into ODS, however.

> For example, assigning operators to runs is done well after the rest of the schedule is finalized.

Can we be clear about what the intended use case is here? I want to make sure I understand who is producing these assignments and who the consumer will be. If the run assignments are done via the same producer after the schedule is finalized but before it would be sent to the schedule consumer, that wouldn't seem like it would impact our decision to include those assignments in the ODS-represented schedule. But I want to be sure that's an accurate read of what you both have said.

skyqrose:
At MBTA, the use case is to export data from HASTUS and distribute it to multiple internal apps using a standard format, instead of the status quo of each app getting a custom export in a custom format. (This data would probably also make it outside of MBTA to Swiftly, but that's not my side of the house so I'm not sure.)

[...]

Issue #30, opened 2023-11-16
add Vehicles.txt 


BTollison:
In an effort to make ODS more compatible with electric fleets, I am proposing the following addition of Vehicles.txt

**Vehicle_type_id**
Used to identify a specific vehicle type within a fleet.

- Required (Primary Key)

**vehicle_long_name**
For the purpose of describing a vehicle type in length

- Example: Where vehicle_type_id could be 1234, a vehicle_description could say "Noord-Holland 10 meter VDL Citea Electric"

**vehicle_short_name**
For the purpose of briefly describing a vehicle.

- Example: vehicle_type_id ias 1234, and vehicle_short_name is "H10e."

**useable_kwh**
Described the amount of useable battery capacity. This should not be the total capacity of the battery as stated by a manufacture, but rather the total allowed consumable and rechargeable kwh. This is a decimal value.

- Example: A manufacture states that your battery is 200kwh total, but that you may only use 80% of that value. Then you are left with 160.00.
- Example 2: Your total battery capacity has degraded from 200kwh to 180kwh, but you are still only allowed to use 80%. Then your useable_kwh should be 144.00.
- This is the primary field for any energy consumption calculation.

**total_kwh**
This represents the full size of the battery as decimal value. It should change over time as batteries degrade.

**charge_curve_id**
Reference to the proposed charge_curves.txt. Connects the appropriate charging curve to the vehicle_type_id.

**min_kwh**
Minimum kwh allowed to remain in the battery while the vehicle is in operation.

- Used to set a bottom threshold of acceptable battery charge level

hodb:
Adding to the vehicles.txt file concept, additional attributes that relate to the vehicle itself (non-EV related) such as - 

- make
- model
- license_plate

These attributes can assist in maintaining fleet information to be consumed by the AVL provider

skyqrose:
Which of these fields would be used to match what's in VehicleDescriptor in GTFS-RT?

BTollison:
The intention of this is actually not to be a database of individual vehicles, but of groups of vehicle types so that up and downstream systems have the information necessary to understand the EV context. I admit the "vehicles.txt" name is a bit misleading in this way.

hodb:
It could be it can be separated into two separated files, connected by a foreign key. My intention was to have communication between the dispatching software and the AVL provider regarding what vehicles are being operated

BTollison:
[...]

PR #81, opened 2024-09-20
potential rosters and employee assignment spec

skyqrose:
[Rendered](https://operational-data-standard-81--cal-itp-previews.netlify.app/spec/#rosterstxt)

Here's an attempt at unifying #28 and #45. It includes both the roster and employee assignments, and both the typical schedule and exceptions. It does not include vehicles.

It includes 4 files:
- `rosters.txt`: A typical week for a roster. Pretty directly corresponds to #45.
- `roster_dates.txt`: Exceptions to `rosters.txt` on specific dates. This makes it possible to describe holidays or vacations that are built into the work, before they're assigned to an employee. It also allows minor disruptions like track work, to say that someone is still doing mostly the same run/roster, but the schedule has changed slightly.
- `employee_rosters.txt`: Which roster did an employee get scheduled to do? For North American schedule, this would refer to most of the rating at once. For UK scheduling, where employees rotate through rosters, there would be one row per employee per week.
- `employee_run_dates.txt`: Exceptions to `employee_rosters.txt`, or allows describing employee assignments without representing rosters, as in #28. If rosters are used, this allows vacations that are scheduled but not considered part of the roster, or leaves of absence.

None of the files require a row for every date + person pair, and it's intended to cover every crew scheduling use case that I remember coming up in working group discussions.

## that's a lot of files

Unfortunately I think it necessarily takes 4 files. To describe what a roster is doing on a date and what an employee is doing on a date requires a `(calendar.txt, calendar_dates.txt)` pair for both rosters and employees. (You could do it in 1 file each by listing every date separately as in #28, but that's duplication that people wanted to avoid, and means you can't describe a typical week.)

I also considered whether we could cut down on the number of files by using `service_ids` so the exceptions could be defined in `calendar_dates.txt` (after all, `run_events` didn't need its own `_dates` file because it could use service IDs), but I don't think that's possible because rosters need to associate a different run with each day.

Vehicles were originally part of the plan in #28. I cut them from the scope of this PR to focus on crew. One concern people had was that if crew required too many files, maybe vehicles will, too. I don't think that's the case, because vehicles don't have human-specific concerns like rosters and vacations. So having 4 crew files doesn't mean it'll get doubled to 8 crew+vehicle files. I think we should make crew rosters first, and then have a separate discussion for vehicles.

## this is a draft

This is just a rough draft to gives something to discuss during working group meetings. It's got some missing pieces and TODOs.

I have a few examples that I've thought about, but don't have them written up yet. If there's any you'd like to see in order to discuss during the working group, let me know, otherwise, I'll put them off til later.


skyqrose:
Some things from the working group meeting to think about / work on:

1. Writing examples would help us be sure that this can represent everything we need to. @BTollison mentioned wanting to do some of this? If you need inspiration, there's a list of examples I think would be useful in [examples.md](https://github.com/cal-itp/operational-data-standard/pull/81/files/f393e8427c316bd61a0d0ac2d6ad03cb266d0484#diff-788373b6538c505c2ade4325b74c888767505552caefabf6b99c5249de77d556R176).
2. Could this lean more on the existing `calendar.txt` and `service_id`s instead of defining its own calendar+date system for these files? I tried and couldn't figure it out, but someone else could give it a shot.
3. When a calendar exception happens (like minor time changes due to track work), you have to copy that exception into `roster_dates.txt` because it refers to runs using a `service_id`. Is there a way to prevent that, and say it still refers to the same run on a different service, without opening the can of worms of having non-unique run ids?
4. Vocabulary: There's a ton of new terms in rostering, which we should define so we're all on the same page.
5. Specifically, this proposal uses `roster_id` to refer to work that one person might do over the course of a week. Usually "Roster" means a bunch of people's schedules, and a single person's schedule might be a "job" or "roster position". What words does your agency use for this? What should we name the files + fields?
6. @BTollison raised a case where agencies have roster positions that last multiple weeks (e.g. a two-week rotation). #45 handled this with a `week_sequence` field. Should we try to add that feature into this proposal, and if so, how?

BTollison:
Alright, I finally (sorry) got to this... here are my thoughts:

In the proposed `rosters.txt` we have `service_id` fields per day and a `start_date` and `end_date`. I think that maybe we can get rid of these. We had discussions earlier about not needing service_id because run_id is tied to a service_id, and I think this idea makes sense now. But to counter a situation where you have multiple sets of rosters or rosters that do not align with the `service_id`, I propose we create a `roster_group.txt`, I'll describe it in more detail below. 

I'd like to propose adding the following to `rosters.txt`

-  `roster_week_sequence` in the case that a roster has multiple weeks within it, for example a 2 week roster. We can number the sequence of the weeks to define which week is week 1 and which is week 2.
- `roster_group_id` ties back to `roster_group.txt`
- `roster_name` this is necessary because `roster_id` is a primary key, but rosters often have specific numbering sequences like blocks or runs do. So, for example it is possible to have a roster 101 that is for the whole of January but then replace it without a new roster set that has a roster 101 starting in February in the same data set.

So in practice you end up with:

**calendar.txt**
<google-sheets-html-origin><style type="text/css"></style>
service_id | monday | tuesday | wednesday | thursday | friday | saturday | sunday | start_date | end_date
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Mon-Wed | 1 | 0 | 1 | 0 | 0 | 0 | 0 | 20240101 | 20240131
Tue-Thur | 0 | 1 | 0 | 1 | 0 | 0 | 0 | 20240101 | 20240131
Friday | 0 | 0 | 0 | 0 | 1 | 0 | 0 | 20240101 | 20240131
Saturday | 0 | 0 | 0 | 0 | 0 | 1 | 0 | 20240101 | 20240131
Sunday | 0 | 0 | 0 | 0 | 0 | 0 | 1 | 20240101 | 20240131

**calendar_dates.txt** -- We replace Monday service with a Holiday.
<google-sheets-html-origin><style type="text/css"></style>
service_id | date | exception_type
-- | -- | --
holiday | 20240101 | 1
Mon-Wed | 20240101 | 2

**roster_group.txt** -- In this example, you can see that we have defined two types of rosters for the same period. Where one is for a 40 hour work week and another is for a 36 hour work week.
<google-sheets-html-origin><style type="text/css"></style>
roster_group_id | start_date | end_date | roster_set_name
-- | -- | -- | --
40_hour_prod | 20240101 | 20240131 | Production
36_hour_prod | 20240101 | 20240131 | Production

- `roster_group_id` defines the primary key per roster set
- `start_date` start date of the roster set
- `end_date` end date of the roster set
- `roster_set_name` human readable name of the roster set

**rosters.txt** -- In this roster we have 3 roster entries, with a 2 week rotation for each, from 2 different sets of rosters that attach to the same `run_id`'s. So, they stay independent of the `service_id'`s but are still linked to them.
<google-sheets-html-origin><style type="text/css"></style>
roster_group_id | roster_id | roster_name | roster_week_sequence | monday_run_id | tuesday_run_id | wednesday_run_id | thursday_run_id | friday_run_id | saturday_run_id | sunday_run_id
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
40_hour_prod | 1 | 101 | 1 | 10 | 10 | 20 | 10 | 10 |   |  
40_hour_prod | 1 | 101 | 2 | 20 | 20 | 10 | 20 |   |   | 210
40_hour_prod | 2 | 102 | 1 | 11 | 11 | 21 |   |   | 111 | 211
40_hour_prod | 2 | 102 | 2 | 21 | 21 |   |   | 21 | 111 | 211
36_hour_prod | 3 | 201 | 1 | 20 |   |   | 20 | 20 | 110 |  
36_hour_prod | 3 | 201 | 2 |   |   | 20 | 10 | 10 | 110 |  

It took me a few days to think about this overall, what do you guys think?

skyqrose:
Oooh thanks for the proposal, there are a ton of good new ideas in here.

1.

> We had discussions earlier about not needing `service_id` because `run_id` is tied to a `service_id`

We had discussions about _trips_ being tied to a `service_id`. But at least at my agency, `run_id`s are re-used between services, so `run_id` does need a `service_id` next to it. It is possible to work around this, but it'd be hard. (More details collapsed cuz it's a bit of a tangent.)
<details>
<summary>How to get unique run IDs</summary>

- Add to the spec that `run_id` must be unique across the whole dataset, like `trip_id` is.
  - (this is backwards incompatible for anyone already using `run_events.txt`, not sure if that applies to anyone yet)
- Translate all our runs from their existing names "500" to new unique names "2024-Fall-weekday-500"
- Have a way to translate "2024-Fall-weekday-500" into the human-readable name "500"
  - Possibly a de-normalized `run_name` column in `run_events.txt`
  - Possibly a new file `runs.txt` with two columns `run_id` and `run_name`.

That's all possible, but in my opinion referring to runs everywhere as a `(service_id, run_id)` pair is easier than renaming the run IDs.
</details>

2. I like the roster group abstraction. This is the first time GTFS or TODS have been able to represent the concept of "The whole spring schedule", and it means we don't have to denormalize the start/end date across every roster in the group. It does mean an extra file, though, not sure if that's worth it.

3. Is roster name meant to be used to match related rosters together by checking which rosters have the same name, or is it just for displaying to humans? Are January 101 and February 101 necessarily related? 

4. How do you know which week of a multi-week roster a date belongs to. Is Wednesday 2024-01-31 part of week 1 or week 2?

5. What is `roster_set_name` used for?

6. The example `roster_dates.txt` edits roster `1`, but not `2` or `3`. This implies that `1` is doing the special holiday run `555`, but `2` and `3` are still doing the normal weekday runs `11` and `20`. Does this proposal solve the problem of needing to add every roster to `roster_dates` even if there's only a minor schedule change (note 3 in my last list)? I was hoping the new `roster_groups` concept could do that somehow but I don't see how. Does run `11` work on trips on both services `Mon-Wed` and `holiday`?

BTollison:
> We had discussions about _trips_ being tied to a `service_id`. But at least at my agency, `run_id`s are re-used between services, so `run_id` does need a `service_id` next to it. It is possible to work around this, but it'd be hard. (More details collapsed cuz it's a bit of a tangent.)

Is this because there is `run_id` for you is the literal run number? I realized this is a problem in the current situation and that we probably need to propose a `run_name` field to allow `run_id` to be a proper primary key.

> 3. Is roster name meant to be used to match related rosters together by checking which rosters have the same name, or is it just for displaying to humans? Are January 101 and February 101 necessarily related?

Yes they can be related in that it's for garage 101 or something like this. I find that in this industry we are often renumbering things in a particular way for a particular business reason. So you are extremely likely to end up with overlapping names, that are meant to be human readable. So it can be useful for looking up as well, especially if you are troubleshooting with someone in the field.

> 4. How do you know which week of a multi-week roster a date belongs to. Is Wednesday 2024-01-31 part of week 1 or week 2?

Yeah, this is something that we can either have a field for or assume that the operations software can control it. I had thought about having a field that basically indicates that week X is the start week of this dataset. So, if you sent an update but didn't change the `start_date` or `end_date`, it's possible to compensate with this field as well. Perhaps we can make this an optional field?

> 5. What is `roster_set_name` used for?

I have seen that often there are multiple roster sets for a garage, so `roster_set_name` can use used to group them together in some way or allow for whatever business related naming convention is necessary in the field. 

> 6. The example `roster_dates.txt` edits roster `1`, but not `2` or `3`. This implies that `1` is doing the special holiday run `555`, but `2` and `3` are still doing the normal weekday runs `11` and `20`. Does this proposal solve the problem of needing to add every roster to `roster_dates` even if there's only a minor schedule change (note 3 in my last list)? I was hoping the new `roster_groups` concept could do that somehow but I don't see how. Does run `11` work on trips on both services `Mon-Wed` and `holiday`?

I lean towards we probably can't solve this because it's hard to see how we aren't stuck trying to tell the programs what needs to happen when we no longer do the normal work that was scheduled. We need to know who does what in place of it. By the way, this is a good time to say I totally forgot about the `employee_rosters.txt`. But Now that I see it again, I think we can just incorporate the employee_id into `rosters.txt` and `roster_dates.txt`

skyqrose:
To break some (not all) of the changes down into smaller topics that could be discussed separately:
- Unique run IDs: Should we refer to runs as `(service_id, non_unique_run_id)`, or should we say that `run_id` must be unique, and add a separate `run_name` field?
- Multi-week rotations: Do we need to support them? If so, then we need to figure out the last details. If not, then agencies who use multi-week rotations would have to reassign people to a different roster position each week.
- Roster groups: Should we add this level of abstraction and add another file? Is normalizing the dates and having a roster set name useful enough to be worth the extra file?

My opinions on them:
- Unique run IDs: Non-unique Run IDs keep causing problems, so I guess it's worth the effort to fix, but it'd probably be backwards incompatible. I'm on the fence.
- Multi-week rotations: It would be nice to represent the data in a way that's more semantic, but I don't see a way to do it without exploding the complexity of the calendar, so I'm leaning opposed to it.
- Roster groups: The benefits to normalizing this data seem small enough now that it doesn't seem worth an extra file, but I could be convinced that it's important.

safrazier17:
@skyqrose I have been having trouble consolidating the proposal because of these questions. Thanks for putting it so neatly. My thoughts:

I don't think we should do another backwards incompatible change at this point. Can you clarify for me: you're saying MBTA would use the same run id twice on a monday and have it assigned to two separate services and containing different trips? How do you differentiate between them in dispatch? Are they for different modes?

multi-week and roster groups: I am inclined to leave out at this point. We are adding a lot of complexity to the spec as is and I think it makes sense to defer these to a future update unless it is actively blocking an in-progress implementation.

jeffkessler-keolis:
Okay, apologies for the delay in weighing in on this.

> Unique run IDs: Should we refer to runs as (service_id, non_unique_run_id), or should we say that run_id must be unique, and add a separate run_name field?

I'm in favor of `run_id` not being forced to a dataset-unique value given the use case of having a consistent `run_id` spanning multiple `service_id`s.

- Those who want to have crew members assigned to a single run regardless of the `service_id`s in effect could do so by leaving the `service_id` definition blank; this is something of which we'd make tremendous use. Others could specify runs based on the specific `servce_id` definition.

- Those of us who have a crew-facing job number can use that figure without needing to use or include an internal identifier, but an internal/unique identifier could always be populated therein.

> Multi-week rotations: Do we need to support them? If so, then we need to figure out the last details. If not, then agencies who use multi-week rotations would have to reassign people to a different roster position each week.

- There are some railroads in the US (e.g. Amtrak) that have positions operate on an alternating cycle (e.g. everyone works one weekend day every-other weekend).

- Many, if not most, foreign operations rely on a months-long roster cycle, whereby the positions are rotated through on a regular basis (and probably common for @BTollison's operations, too).

- I see a few ways this approach could be implemented:

	- Every multi-week roster is itemized out with weekly roster dates, each of which have to be assigned to their specific dates.
	- The roster is listed with each week in succession, with the start listed in a date-based assignment. (This is still the same amount of data if over a long period of tine, but the date-interpretation becomes implicit vs explicit, and could be repeated for multi-week cycles.)
	- The roster is listed with each week in succession, and individuals could be assigned to both a roster position AND a starting week. (This reduces the amount of data required to convey the applicable information, and although implicit with dates, is explicit in operation and intentionality.)


> Roster groups: Should we add this level of abstraction and add another file? Is normalizing the dates and having a roster set name useful enough to be worth the extra file?

I'm admittedly not sure how much is needed here / not sure I get the use case. Happy to have this abstraction as an optional grouping, but I don't know if I am understanding a reason to make it a mandatory structural element at this point.

safrazier17:
Summary of decisions from today's meeting:

* We will add a roster_groups file that will not include start and end date, which will instead remain in the rosters.txt file (@skyqrose, can you incorporate this into the PR?)
* We will defer consideration of multi-week rosters; this is likely a high priority for Euro operators, but adds a level of complexity that we are not currently prepared to consider in v2.1
* No final decision has been made on run_id uniqueness, but we are leaning toward continue to key based on the unique run_id/service_id pair rather than introducing a new run_name concept
* @antrim will build a vehicle assignments proposal based on @BTollison's #30; will report on whether this is a blocker for current stage implementation

skyqrose:
Notes from last Monday's meeting:
- We won't include roster groups. (Reasons: It only gives two new fields: roster_group_id, and roster_set_name. The purpose isn't clear, they'll be easy to add later, and we want to keep things as small as possible for now. But Bradley, you weren't there to say why it's important, so if it's really important to you, you can still push back on this.)
  - (I just had a new idea: What if `rosters.txt` had an optional column `roster_set_name`? That might get most of the benefit of roster groups without requiring an extra file.)
- We won't include multi-week rosters. (We know it's important, but it's very complex, we don't have a good solution, and the data is still possible to represent week-by-week.)
- I'm working on some ideas on run_id uniqueness and service IDs. Jeff wants service IDs to be optional in order to avoid redoing roster assignments when there are small changes to a schedule. May or may not be possible, still investigating, but I'm hopeful for some significant improvements on this front.

Other work on finishing this PR (edits to the spec, writing more docs, examples) is blocked on that run_id investigation.

skyqrose:
Okay, I've edited the spec to make service ID optional, according to Jeff's recommendation.

I've also cleaned up some other stuff. Everything should be complete and ready for another round of feedback, except for the examples (which I'm still working on).

antrim:
Thanks @skyqrose Should we use this comparison view to see the changes?
https://github.com/cal-itp/operational-data-standard/compare/fb4e3b3a03b93ac1d365901904d3419875c7c725..f393e8427c316bd61a0d0ac2d6ad03cb266d0484

skyqrose:
I'm done writing examples, and this is ready for another review from the working group and from @antrim 's coworkers.

Changes since the last meeting:
- Examples, including pseudocode algorithms for looking up data through all the calendar files.
- Moved `employee_run_dates` to the top, so it's the primary way to use this data, and the roster position files are more like optional extras.
- Minor change: forbid specifying the run if exception_type=2, cuz that felt unnecesary when writing the examples.

skyqrose:
But also, this is a lot of complexity. Now that we have full examples to look at, in our next working group meeting we should discuss whether this is all worth it. I have another branch that does a much simpler thing: `employee_run_dates.txt` only, service IDs required. (I'll edit this post with links coming in a few minutes.)

Reasons to have this complex version:
- It can represent the concept of rosters, and a regular weekly schedule. These concepts don't exist at all in the simpler version, there's no alternative or workaround.
- Without rosters to represent regular data, `employee_run_dates` would get huge. Number of employees * number of days.
- It would be difficult (backwards-incompatible) to add those concepts back later if we start with assuming the whole schedule is in `employee_run_dates.txt`.

Reasons to do the simple version:
- Spec is way easier to understand. 1 simple file instead of 4 mediumly complex files.
- Consuming data is waaaay easier. Going through every layer of calendars and exceptions makes the simple question of "who's working today" 10 times more complex.
- Do we even have stakeholders who care about rosters?
    - Optibus represents rostering completely internally, so when exposing data to customers was planning to only use `employee_run_dates.txt` anyway.
    - MBTA subway and bus operations mostly just care about who's driving each vehicle. It doesn't matter whether they're a 
    - I think I remember Jeff and Bradley expressing more support for rosters? Now that things are more finalized, I'd like to hear from them again to see if they think the added features are worth the complexity.
- In general, GTFS is focused on representing the service that's happening, the final output of the scheduling process, as opposed to representing internal details like target headways or how employees pick work. You could argue that how work is grouped into roster positions is an internal concern of the scheduling process, and once they've been assigned to runs, the roster positions don't matter anymore.

This is probably too much to discuss in GitHub comments, so no pressure to reply to this specifically, I'll bring it up again during the working group meeting.

docs/spec/index.md
```diff
@@ -139,3 +145,98 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+
+Employees are assigned to these roster positions in [`employee_roster.txt`](#employee_rostertxt).
+
+Primary Key: `roster_position_id`
+
+| **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- |
+| `roster_position_id` | Unique ID | Required | Unique within dataset |
+| `start_date` | Date | Required | First service day that the roster position works. |
+| `end_date` | Date | Required | Last service day tha the roster position works. This day is included in the interval. |
```
jfabi, line 181:
```suggestion
| `end_date` | Date | Required | Last service day that the roster position works. This day is included in the interval. |
```

docs/spec/index.md
```diff
@@ -139,3 +145,98 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+After evaluating [`roster.txt`](#rostertxt) and `roster_dates.txt`, each run can only be assigned to one roster position on each date. A roster position may be scheduled to do multiple runs on the same date.
+
+This file may be used even when [`roster.txt`](#rostertxt) is not defined, in which case each roster position is made up of the dates added in this file. This may be useful for agencies whose rosters are very irregular. In this case, the `exception_type` column can be omitted because every row is adding a date, which is the default when the column is blank.
+
+Primary Key: `*`
+
+| **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- |
+| `roster_position_id` | ID referencing [`roster.txt`](#rostertxt) or ID | Required | If `exception_type` is `1`, then the ID does not have to appear in [`roster.txt`](#rostertxt). This file may define new roster positions. |
+| `date` | Date | Required | Date when exception occurs. |
```
jfabi, line 224: Should this explicitly refer to service date?

docs/spec/index.md
```diff
@@ -139,3 +145,98 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+If [`employee_roster.txt`](#employee_rostertxt) is used, then describes exceptions to that schedule.
+
+If a feed doesn't represent roster positions, it can still assign employees to runs by putting every run for every date in this file. In that case, the `exception_type` column can be omitted because every row would be adding a date, which is the default when the column is blank. [Example](./examples/rostering.md#simplest-example-employee_run_datestxt-only).
+
+Primary Key: `*`
+
+| **Field Name** | **Type** | **Required** | **Description** |
+| --- | --- | --- | --- |
+| `date` | Date | Required | |
+| `employee_id` | ID | Required | References an agency's external systems. Employee IDs are not used elsewhere in TODS. |
```
jfabi, line 162: They are also used in `employee_roster.txt`, which is implied in the file description above, but the phrasing of this note sounds a bit awkward in light of that.

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+## Simplest example: employee_run_dates.txt only
+
+The simplest way to assign employees to runs is to use [`employee_run_dates.txt`](/docs/spec.md#employee_run_datestxt). This will require one row per employee per date.
+
+In this example, `A` and `B` work Saturday Feb 1 and Wednesday Feb 5 through Friday Feb 7. `C` and `D` work Sunday Feb 2 through Tuesday Feb 4.
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 14: missing `sunday` in field names

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+In this example, the roster includes holidays, which are defined as exceptions in [`roster_dates.txt`](/docs/spec.md#roster_datestxt).
+
+This agency has two workers, who work M-F. There's no service on weekends. On each holiday, the agency runs half the service, and each worker gets one holiday off.
+
+The holidays are built into the roster positions, so there's no need for [`employee_run_dates.txt`](/docs/spec.md#employee_run_datestxt).
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 269: missing `sunday` in field names

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+In this example, this agency runs service 4 days a week, with two employees who each regularly work two days per week. When each employee goes on vacation, their work is covered by a third employee on a third roster position.
+
+These vacations are built in to the roster.
+
+The third employee has no regular work, so appears in [`roster_dates.txt`](/docs/spec.md#roster_datestxt) but not [`roster_positions.txt`](/docs/spec.md#roster_positionstxt).
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 337: missing `sunday` in field names

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+## Multi-week roster positions
+
+At some agencies (especially in Europe), roster positions repeat every two weeks. In TODS, this can be represented by having the first and second weeks be two different roster positions, and using [`employee_roster.txt`](/docs/spec.md#employee_rostertxt) to assign employees to each roster position each week.
+
+In this example, there's one run per day on weekdays only. Position A works Mondays and Tuesdays. Position B works Thursdays and Fridays. They trade off on Wednesdays, so each position works an average of 2½ days per week.
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 442: missing `sunday` in field names

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+## Rotating assignments
+
+At some agencies (especially in the UK), employees rotate among all roster positions over many weeks. In TODS, this can be represented by assigning each employee to a new roster position each week.
+
+In this example, there are 5 employees and 5 roster positions. Over a calendar of 3 weeks, each employee will work 3 of the 5 positions.
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 479: missing `sunday` in field names

docs/spec/examples/rostering.md
```diff
@@ -0,0 +1,568 @@
...
+If the roster files specify service IDs (which is recommended), then either [`roster_dates.txt`](/docs/spec.md#roster_datestxt) or [`employee_run_dates.txt`](/docs/spec.md#employee_run_datestxt) must be used to assign the roster position or employee to the new `(service_id, run_id)` pair, i.e. `(trackwork, 100)` instead of `(weekday, 100)`.
+
+However, in this example, the roster does not specify the service ID in [`roster_positions.txt`](/docs/spec.md#roster_positionstxt). Therefore, the employee is assigned to run 100 on whichever service is active. Since both the regular and replacement runs have run ID `100`, on the day of the exception, the employee will be implicitly assigned to `(trackwork, 100)` without needing to specify that in the roster files.
+
+The spec describes this situation in [Service IDs in Rosters](/docs/spec/index.md#service-ids-in-rosters).
+
+**[`calendar.txt`](https://gtfs.org/documentation/schedule/reference/#calendartxt)**
+
+```csv
+service_id,monday,tuesday,wednesday,thursday,friday,saturday,start_date,end_date
```
safrazier17, line 528: missing `sunday` in field names

skyqrose:
After discussion with the working group, closing this PR in favor of #87 . Scott will send out more info about this in an email today.

PR #78, opened 2024-08-13
Rename ods to tods

isabelle-dr:
ODS was [renamed to TODS](https://ods.calitp.org/governance/actions/) in June 2024.
This PR renames ODS to TODS so it appears everywhere on the website and there is no confusion.

PR #83, opened 2024-11-19
spelling and grammer edits

vykster:


safrazier17:
@vykster these are in #78

PR #85, opened 2024-12-13
Vehicle assignments

antrim:
Proposed spec in response to issue #28 "Allocating drivers and vehicles in ODS".

**What it does:** Assign individual vehicles to blocks.

**What it does NOT do:** 
- Introduce vehicle types, categories, groupings, or couplings. These could be added in a future version, looking towards drawing from the [http://bit.ly/gtfs-vehicles](GTFS-Vehicles proposal).
- Assign vehicles according to a regular weekly schedule (i.e. instead of individual service date)

**Changes included:**
- `vehicles.txt`: ID, text label, license plate. Only a few attribute fields are defined for a minimum viable proposal. TODS consumers may be able to look up additional attributes by vehicle_id. More attributes can be defined in future TODS versions. 
- `vehicle_assignments.txt`: Assignments by service date, `block_id`, `service_id`


antrim:
Based on this https://github.com/cal-itp/operational-data-standard/issues/28#issuecomment-2292345857 -  what @skyqrose outlined in August

Notes:

- vehicles.txt and vehicle_categories.txt crib liberally from http://bit.ly/gtfs-vehicles. That proposal aligned with platform descriptions for GTFS-Pathways. I don't know the current status of that proposal. I included a baseline of fields; perhaps more should be included in this proposal.
- I considered whether to relate vehicle assignments to run_events.txt, but since mid-run events would never involve the change of a vehicle, this would have created more complexity.
- vehicle_assignments can map to `trip_id` or `block_id`. This is for cases where block_id is not universally supplied, perhaps because a GTFS feed follows the original usage of `block_id` - indicating allowed in-seat transfers.

antrim:
Some internal discussion at Optibus:
Timon Kelter suggested that `vehicle_categories` sounds very abstract, but `vehicle_types` would be more natural and is descriptive. I agree. I had was cribbing from the GTFS-Vehicles proposal, but don't see a reason to keep this term. Will plan to change to `vehicle_types.txt`.
BTW, I have invited Optibus team to share more discussion here. So we'll expose a bit more deliberation that might otherwise have been internal :)

antrim:
These are vehicle properties that Optibus currently exports in a non TODS format. Which properties should be included in TODS?

[Vehicle properties example.xlsx](https://github.com/user-attachments/files/18122907/Vehicle.properties.example.xlsx)


| vehicle_id | ID | Identifies a vehicle. |
|---|---|---|
| agency_id | ID referencing agency.agency_id | Agency for the specified vehicle. |
| start_date | Date | Day when the vehicle begins to be part of the fleet. |
| license_plate | Text | License plate of the specified vehicle. |
| make | Text | Make of the specified vehicle. |
| model | Text | Model of the specified vehicle. |
| owner | Text | Vehicle owner entity. |
| registration_date | Date | Date of the first registration. |
| available_seats | Non-negative Integer | Number of seating places. |
| available_standing | Non-negative Integer | Number of standing places. |
| typology | Enum | Vehicle typology. <br/>0 or empty = unknown typology <br/>1 = Urban Mini <br/>2 = Urban Midi <br/>3 = Urban Standard <br/>4 = Urban Articulated <br/>5 = Inter-urban Standard <br/>6 = Inter-urban Articulated <br/>7 = Tourism |
| vclass | Enum | Vehicle class. <br/>0 or empty = unknown class <br/>1 = A (Urban Mini) <br/>2 = I (Urban Midi, Urban Standard, Urban Articulated) <br/>3 = II (Inter-urban Standard, Inter-urban Articulated) <br/>4 = III (Tourism) |
| propulsion | Enum | Propulsion technology. <br/>0 or empty = unknown propulsion <br/>1 = Gasoline <br/>2 = Diesel <br/>3 = LPG auto <br/>4 = Mixture <br/>5 = Biodiesel <br/>6 = Electricity <br/>7 = Hybrid <br/>8 = Natural Gas <br/>9 = Other |
| emission | Enum | Emissions class. <br/>0 or empty = unkwon emissions class <br/>1 = EURO I <br/>2 = EURO II <br/>3 = EURO III <br/>4 = EURO IV <br/>5 = EURO V <br/>6 = EURO VI |
| new_seminew | Enum | New or Semi-New. <br/>0 or empty = no <br/>1 = yes |
| ecological | Enum | Non-polluting or energy-efficient vehicle. <br/>0 or empty = no <br/>1 = yes |
| climatization | Enum | Climatization. <br/>0 or empty = no <br/>1 = yes |
| wheelchair | Enum | Wheelchair accessible. <br/>0 or empty = no <br/>1 = yes |
| corridor | Enum | Passageway for wheelchairs / baby strollers. <br/>0 or empty = no <br/>1 = yes |
| lowered_floor | Enum | Lowered floor. <br/>0 or empty = no <br/>1 = yes |
| ramp | Enum | Ramp for people with reduced mobility. <br/>0 or empty = no <br/>1 = yes |
| folding_system | Enum | Folding system for boarding and exiting of people with reduced mobility. <br/>0 or empty = no <br/>1 = yes |
| kneeling | Enum | Vehicle "kneeling" to reduce entry height. <br/>0 or empty = no <br/>1 = yes |
| static_information | Enum | Posters and fixed signage. <br/>0 or empty = no <br/>1 = yes |
| onboard_monitor | Enum | On-board monitor. <br/>0 or empty = no <br/>1 = yes |
| front_display | Enum | External front destination display. <br/>0 or empty = no <br/>1 = yes |
| rear_display | Enum | External rear destination panel. <br/>0 or empty = no <br/>1 = yes |
| side_display | Enum | External right side destination panel. <br/>0 or empty = no <br/>1 = yes |
| internal_sound | Enum | Internal sound information (on-board Public Address). <br/>0 or empty = no <br/>1 = yes |
| external_sound | Enum | External sound information (external Public Address). <br/>0 or empty = no <br/>1 = yes |
| consumption_meter | Enum | Consumption measurement. <br/>0 or empty = no <br/>1 = yes |
| bicycles | Enum | Permission to transport non-folding bicycles (folding bicycles are always allowed). <br/>0 or empty = no <br/>1 = yes |
| passenger_counting | Enum | Vehicle has a passenger counting system. <br/>0 or empty = no <br/>1 = yes |
| video_surveillance | Enum | Vehicle has a video surveillance system. <br/>0 or empty = no <br/>1 = yes |

timon-k:
From the fields mentioned above, I would remove at least `new_seminew`, `ecological`,  `static_information`, `onboard_monitor`, all `*_display` fields, all `*_sound` fields and `consumption_meter`.

To me, those look too undefined, covered by other attributes like `emissions` already or just not needed for operational purposes.

I'm also very unsure about `folding_system`, `ramp` and `kneeling`. Given that we have `wheelchair` and `lowered_floow` already that looks like too many details on the same question.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+|---|---|---|---|
+| `date` | Date | Required |  |
+| `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | Optional | Identifies a set of dates when the run is scheduled to take place. Required if `block_id`s are repeated between different `service_id`s. |
+| `block_id`		  | ID referencing `trips.block_id` | Conditionally required | Identifies the block. Either `trip_id` or `block_id` must be specified. |
+| `trip_id` | ID referencing `trips.trip_id` | Conditionally required | Either `trip_id` or `block_id` must be specified. In the case where both are supplied, `trip_id` overrides `block_id`. Note: multiple vehicles are allowed on the same block+date, but this is not recommended. |
+| `vehicle_id` | ID referencing `vehicles.vehicle_id` | Conditionally required | Refers to a specific vehicle in the transit fleet. Either `vehicle_id` or `vehicle_category_id` MUST be supplied. |
+| `vehicle_category_id` | ID referencing `vehicle_categories.vehicle_category_id` | Conditionally required | Refers to a category of vehicle in the transit fleet if there is no specific vehicle assignment. Either `vehicle_id` or `vehicle_category_id` MUST be supplied. |
```
timon-k, line 228: Having the category here sounds good, but perhaps we should then stress that this is the _operationally assigned_ category of vehicle, not the planned one.

The planned vehicle category/type would be more naturally placed in `routes_supplement.txt` or `trips_supplement.txt`. Usually, planned and real vehicle type should be identical, but there could be differences.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `wheelchair_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders in a wheelchair that the vehicle can carry. |
+
+### `vehicles.txt`
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. |
+| `vehicle_ category_id` | ID referencing `vehicle_categories.vehicle_category_id` | Optional |  |
```
timon-k, line 253: Small typo here, should be `vehicle_category_id`

antrim: @timon-k : Thanks. Resolved

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `vehicle_description` | Text | Optional |  |
+| `registration_date` | Date | Optional |  |
+| `license_plate` | Text | Optional | License number of the vehicle for identification, e.g. “E898656” |
+| `owner` | Text | Optional | Registered owner, e.g “City of Arcata” |
```
skyqrose, line 260: Perhaps `owner` should be part of the vehicle type instead. If two owners own similar vehicles in the same feed, it's okay if they're represented as two types, cuz they're probably not interchangeable anyway.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+### `vehicles.txt`
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. |
+| `vehicle_ category_id` | ID referencing `vehicle_categories.vehicle_category_id` | Optional |  |
+| `vehicle_description` | Text | Optional |  |
+| `registration_date` | Date | Optional |  |
```
skyqrose, line 258: What should be the reasoning for which fields belong in TODS?

`registration_date` doesn't seem that useful for assigning vehicles to trips.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `license_plate` | Text | Optional | License number of the vehicle for identification, e.g. “E898656” |
```
skyqrose, line 259: There may need to be a `label` field in addition to `license_plate`, similar to [GTFS-RT VehicleDescriptor](https://gtfs.org/documentation/realtime/feed-entities/vehicle-positions/#vehicledescriptor).

docs/spec/index.md
antrim: Yes. [GTFS-Vehicles]([bit.ly/gtfs-vehicles](http://bit.ly/gtfs-vehicles)) had proposed to add a `routes`.`vehicle_category_id` field that would define "a default vehicle category for all trips belonging to this route". This could be useful for rider-facing applications (i.e. accessibility & amenity information) and for planning, scheduling, and operational purposes as well, I imagine.

docs/spec/index.md
```diff
@@ -139,3 +141,120 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+Each run can only be assigned to one employee on each date. Employees may be scheduled to more than one run on the same date.
+
+### `vehicle_assignments.txt`
+
+Primary Key: `*`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `date` | Date | Required |  |
+| `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | Optional | Identifies a set of dates when the run is scheduled to take place. Required if `block_id`s are repeated between different `service_id`s. |
```
skyqrose, line 224: description mentions runs, probably a copy-paste error.

How do date and service_id interact? Why do you need service_id when you already have a date? Are block_ids unique within the service but not the date?

antrim:
@skyqrose @safrazier17 Can you help guide process here? Since I worked off the #81 branch, this has the rosters and employee assignments spec draft. Should I remove those changes?

Will we vote on the two PRs separately or together? Merge separately?

docs/spec/index.md
```diff
@@ -139,3 +141,120 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `block_id`		  | ID referencing `trips.block_id` | Conditionally required | Identifies the block. Either `trip_id` or `block_id` must be specified. |
+| `trip_id` | ID referencing `trips.trip_id` | Conditionally required | Either `trip_id` or `block_id` must be specified. In the case where both are supplied, `trip_id` overrides `block_id`. Note: multiple vehicles are allowed on the same block+date, but this is not recommended. |
```
skyqrose, line 226: Would an agency ever schedule multiple vehicles to the same block? Doesn't that go against the definition of block? I know that might happen as a realtime adjustment, but if it's scheduled to happen, shouldn't they change the schedule so it's multiple different blocks?

If we can make that assumption, then this field could be removed.

docs/spec/index.md
```diff
@@ -139,3 +141,120 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+
+### `vehicle_assignments.txt`
+
+Primary Key: `*`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `date` | Date | Required |  |
+| `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | Optional | Identifies a set of dates when the run is scheduled to take place. Required if `block_id`s are repeated between different `service_id`s. |
+| `block_id`		  | ID referencing `trips.block_id` | Conditionally required | Identifies the block. Either `trip_id` or `block_id` must be specified. |
```
skyqrose, line 225: Block could be always required if:
- We say you can only use this feature if you use block IDs (they're optional in [trips.txt](https://gtfs.org/documentation/schedule/reference/#tripstxt))
- Either:
    - You have to include the block, even if you include the trip. (And the block must be the same block as the trip's block.)
    - trip_id is removed (see previous comment).

Alternatively, it would also be possible to say that exactly one of block_id and trip_id is required, which might be less error-prone than allowing both and ignoring the block if the trip exists.

docs/spec/index.md
```diff
@@ -139,3 +141,120 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `trip_id` | ID referencing `trips.trip_id` | Conditionally required | Either `trip_id` or `block_id` must be specified. In the case where both are supplied, `trip_id` overrides `block_id`. Note: multiple vehicles are allowed on the same block+date, but this is not recommended. |
+| `vehicle_id` | ID referencing `vehicles.vehicle_id` | Conditionally required | Refers to a specific vehicle in the transit fleet. Either `vehicle_id` or `vehicle_type_id` MUST be supplied. |
+| `vehicle_type_id` | ID referencing `vehicle_types.vehicle_type_id` | Conditionally required | Refers to a type of vehicle in the transit fleet if there is no specific vehicle assignment. Either `vehicle_id` or `vehicle_type_id` MUST be supplied. |
```
skyqrose, line 228: Can both `vehicle_id` and `vehicle_type_id` be specified? If so, then the spec should specify that the vehicle must have that type.

docs/spec/index.md
```diff
@@ -139,3 +141,120 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `run_id` | ID referencing `run_events.txt` | Required | The run that's either added or removed from this employee's schedule. If `exception_type` is `2` and `run_id` is not blank, then it must match a Run ID that the employee was scheduled to do on this date according to `employee_rosters.txt`, `rosters.txt` and `roster_dates.txt`. |
+| `exception_type` | Enum | Optional | `1` (or blank) - The run is assigned to this employee on the specified date.<br />`2` - The employee will not work this run on this date. |
+
+If a feed doesn't represent rosters, it can still assign employees to runs by putting every run for every date in this file. In that case, the `exception_type` column can be omitted because every row would be adding a date, which is the default when the column is blank.
+
+Each run can only be assigned to one employee on each date. Employees may be scheduled to more than one run on the same date.
+
+### `vehicle_assignments.txt`
+
+Primary Key: `*`
```
skyqrose, line 219: This key allows multiple vehicles/types to be assigned to the same block/trip. Is that intentional, as a way to allow any of the vehicles to run the trip? If so, then that should be mentioned in the docs. If not, then a primary key that would prevent that is `(date, block_id, trip_id)`.

skyqrose (commented):
I've left lots of line-by-line comments.

As for the process, I think this is completely independent from #81. We could split them into two separate PRs/votes.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `block_id`		  | ID referencing `trips.block_id` | Conditionally required | Identifies the block. Either `trip_id` or `block_id` must be specified. |
+| `trip_id` | ID referencing `trips.trip_id` | Conditionally required | Either `trip_id` or `block_id` must be specified. In the case where both are supplied, `trip_id` overrides `block_id`. Note: multiple vehicles are allowed on the same block+date, but this is not recommended. |
+| `vehicle_id` | ID referencing `vehicles.vehicle_id` | Conditionally required | Refers to a specific vehicle in the transit fleet. Either `vehicle_id` or `vehicle_category_id` MUST be supplied. |
+| `vehicle_category_id` | ID referencing `vehicle_categories.vehicle_category_id` | Conditionally required | Refers to a category of vehicle in the transit fleet if there is no specific vehicle assignment. Either `vehicle_id` or `vehicle_category_id` MUST be supplied. |
+
+Not every trip or block and date combo needs to have a vehicle specified.
+
+### `vehicle_categories.txt`
```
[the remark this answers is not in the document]
antrim: It would make sense for vehicle_categories (now `vehicle_types.txt`) to live outside of TODS. However, I do not believe that it has been implemented yet. So, I think we should implement in TODS in a way that could become part of a "primary GTFS" dataset later.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+
+Primary Key: `vehicle_category_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_category_id` | ID, primary key | Required | Defines an ID for a vehicle category. |
+| `vehicle_category_name` | Text | Optional | The vehicle_category_name field defines the name of the vehicle category.   E.g. “MR73” in Montréal, “TGV Duplex” in France or “8-car Waratah Train” in Sydney. |
+| `fuel_type` | Enum | Optional | 0 or empty - unknown propulsion <br />1 - Gasoline <br /> 2 - Diesel <br /> 3 - LPG auto <br /> 4 - Mixture <br /> 5 - Biodiesel <br /> 6 - Electricity <br /> 7 - Hybrid <br /> 8 - Natural Gas <br /> 9 - Other |
```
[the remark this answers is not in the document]
antrim: Also, this list is messy -- properties are mixed up. "Hybrid" isn't a fuel type. For now, I favor removing or replacing with a text string if it's necessary for an immediate use case. (But imagining that the vehicle_category_name could make useful distinctions, e.g. `vehicle_category_name`="40' LNG Bus".

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `wheelchair_accessible` | Enum | Optional | Wheelchair accessible. <br />0 or empty - no <br />1 - yes |
+| `seating_capacity` | Non-negative Integer | Optional | This number denotes the number of seats dedicated to riders, excluding folding seats. A seat is considered accommodating only one rider in a seated position. |
+| `max_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders that the vehicle can carry. |
+| `wheelchair_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders in a wheelchair that the vehicle can carry. |
+
```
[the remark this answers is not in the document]
antrim: At this point, I think we should implement only fields that will be consumed by some software either out-of-the-gate or soon. Any other fields discussed could be implemented when there is a use later.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+### `vehicles.txt`
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. |
```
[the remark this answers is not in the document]
antrim: There is theoretical usefulness of a consistent `vehicle_id`. So the spec might make a recommendation. How shall we assess this? Ask reps from Transit, Swiftly, GMV, transit agencies or others to say whether this makes sense, offers value?

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `vehicle_category_id` | ID, primary key | Required | Defines an ID for a vehicle category. |
+| `vehicle_category_name` | Text | Optional | The vehicle_category_name field defines the name of the vehicle category.   E.g. “MR73” in Montréal, “TGV Duplex” in France or “8-car Waratah Train” in Sydney. |
+| `fuel_type` | Enum | Optional | 0 or empty - unknown propulsion <br />1 - Gasoline <br /> 2 - Diesel <br /> 3 - LPG auto <br /> 4 - Mixture <br /> 5 - Biodiesel <br /> 6 - Electricity <br /> 7 - Hybrid <br /> 8 - Natural Gas <br /> 9 - Other |
+| `wheelchair_accessible` | Enum | Optional | Wheelchair accessible. <br />0 or empty - no <br />1 - yes |
+| `seating_capacity` | Non-negative Integer | Optional | This number denotes the number of seats dedicated to riders, excluding folding seats. A seat is considered accommodating only one rider in a seated position. |
+| `max_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders that the vehicle can carry. |
+| `wheelchair_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders in a wheelchair that the vehicle can carry. |
+
+### `vehicles.txt`
```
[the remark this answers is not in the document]
antrim: So, this is one of the reasons that I have proposed to build off GTFS-Vehicles. The draft spec has something called [GTFS-VehicleCouplings](https://docs.google.com/document/d/156RiBjI6FnWJvO8_XWX11Q9nLdOiBdS_rilA-oamlv8/edit?tab=t.0#bookmark=id.ouilwc42x9co) that answers this.

docs/spec/index.md
```diff
@@ -139,3 +141,117 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. |
+| `vehicle_ category_id` | ID referencing `vehicle_categories.vehicle_category_id` | Optional |  |
+| `vehicle_description` | Text | Optional |  |
```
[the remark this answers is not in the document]
antrim: Was assuming that descriptive text strings can end up being useful for edge cases we can't anticipate.

docs/spec/index.md
antrim: > description mentions runs, probably a copy-paste error.

Oops, will fix.

> Why do you need service_id when you already have a date?

Service periods defined by calendar.txt can overlap. E.g. Monday-Sunday with additional Monday-Friday only trips.

docs/spec/index.md
antrim: I included `trip_id` as an option for cases where `block_id` is not supplied in the GTFS. So, on 2nd thought I'd favor removing, unless we know this is a case to plan for out of the gate.

docs/spec/index.md
antrim: Let's explore removing `vehicle_assignments`.`trip_id`

docs/spec/index.md
antrim: Hmm… I'll add a primary key. `(date, block_id, service_id)`.
Allowing multiple assignments or planning for couplings seems like a complex case.

docs/spec/index.md
antrim: aee1524

docs/spec/index.md
antrim: f3281fe

docs/spec/index.md
antrim: b4c4be1

docs/spec/index.md
antrim: f3281fe536c40490a0fb9a221c7cfd1f41b64e15

docs/spec/index.md
antrim: 85f67e58a8a68ed06d1b0e1ee4ffbd55d3e294ba

docs/spec/index.md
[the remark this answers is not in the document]
jeffkessler-keolis: Is there a reason to exclude folding seats from the `seating_capacity` field? (Is the idea that it is the max number of seats inclusive of wheelchairs in one vehicle?)

I know there are many agencies that almost exclusively use transverse folding seats, which could lead to artificially-lower `seating_capacity` values.

docs/spec/index.md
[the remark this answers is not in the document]
jeffkessler-keolis: I see this field being useful for vehicle-specific things, even more esoteric items like "Veterans-Wrapped Coach" or "Heritage Paint Scheme."

docs/spec/index.md
jeffkessler-keolis: @skyqrose Yes and no. In interstate rail, it's not uncommon for various identical pieces of equipment to be owned by different operators (e.g. 4/120 SEPTA SLV cars are owned by the State of Delaware, but are otherwise identical and interchangeable with the rest of the fleet; MNR's M8s are owned in a mix between MNR and ConnDOT, but the cars are likewise identical). From a planning perspective, we wouldn't necessarily care _whose_ car was operated, just that one of those cars was operated along the trip.

docs/spec/index.md
jeffkessler-keolis: Minor thing (assuming we're in agreement): we should probably clarify that it's `service_date`, not just the date of assignment or the trip's calendar date of operation.

docs/spec/index.md
jeffkessler-keolis: I'd recommend including it for specific trips. This works for agencies that don't publish blocks and may not wish to make specific trip-specific block assignments in TODS, and likewise works for agencies with interchangeable operations. (I could see us making these assignments on a trip-specific level to specify requirements by trip prior to building out cycles.)

timon-k: I'm in favour of keeping it as it is now, with the `trip_id` removed. 

If we allow assignments of individual trips to vehicles it breaks the logic of blocks as @skyqrose pointed out.

If we allow assignments of individual trips to vehicle _types_/_categories_ it overlaps content-wise with http://bit.ly/gtfs-vehicles (which already has a route-to-vehicle-category assignment modeling a "requirement"). The case that you pointed out last ("specify requirements by trip prior to building out cycles") should probably be modeled using http://bit.ly/gtfs-vehicles.

"agencies that don't publish blocks" can still wrap each trip in an artificial block and assign it this way and/or add their real blocks in the non-public `trips_supplement.txt` only. 

So overall, it seems like we can transmit the operational plan (including which vehicles/types will really be running the trips) without allowing to include `trip_id` in the assignments.

docs/spec/index.md
antrim: @jeffkessler-keolis Good point. I'll take a closer look at the GTFS definitions around service days.

antrim:
We've been discussing this internally at Optibus, including with some CAD/AVL partners. For the upcoming version, we propose a "least viable" approach:

- Remove `vehicle_types.txt`
- Scale `vehicles.txt` down to `vehicle_id` and `license_plate`

This removes speculative vehicle attributes. The CAD/AVL should be able to look up vehicle attributes stored in that system.


Thoughts/reactions?

skyqrose:
That does remove the ability to assign a type of vehicle without assigning a specific vehicle, but that's probably fine if we're aiming for a least viable approach.

antrim:
@skyqrose If we pass a version of this without vehicle types, I think the discussion was still useful, so thanks for the review and great comments. And if we do add in more vehicle attributes besides the ID and license plate, then bringing back in vehicle type seems useful.

docs/spec/index.md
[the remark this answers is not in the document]
skyqrose: From working group meeting, Jeff recommends not forcing this ID to match the GTFS-RT ID, because planning and realtime data use different systems. E.g. he has realtime data only for the cab car, but planning system plans every car in the consist.

We can recommend that it's good to keep them in sync, but not guaranteed.

docs/spec/index.md
antrim: Agree. I will remove.

docs/spec/index.md
antrim: b3d2bec

safrazier17:
Pre-vote checklist (@antrim)

- [ ] Remove `vehicle_types.txt`
- [ ] Remove `vehicles.vehicle_type_id`
- [ ] Remove `vehicles.vehicle_description`
- [ ] Remove `vehicles.registration_date`
- [ ] Remove `vehicles.owner`
- [ ] Edit description of `vehicles.license_plate` to account for its use as a more generic unique vehicle identifier
- [ ] Add `vehicles.vehicle_label`
- [ ] Remove `vehicle_assignments.vehicle_type_id`
- [ ] Edit `vehicle_assignments.vehicle_id` description to include note about roadmap plan to allow assignments to be made to groups of vehicles (a la vehicle type) rather than just to individual vehicles
- [ ] Add vehicle assignments examples
- [ ] Edit the PR description so that it includes an overview of proposed changes as they currently stand and the goal of the changes

docs/spec/index.md
```diff
@@ -149,37 +149,20 @@ Primary Key: `(date, block_id, service_id)`
...
-| `vehicle_description` | Text | Optional |  |
-| `registration_date` | Date | Optional |  |
-| `license_plate` | Text | Optional | License number of the vehicle for identification, e.g. “E898656” |
-| `owner` | Text | Optional | Registered owner, e.g “City of Arcata” |
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. It is *recommended* but not required to match the `vehicle_id` in GTFS-realtime feeds. |
+| `vehicle_label` | Text | Optional | Free text label for a vehicle, e.g. vessel name. |
+| `license_plate` | Text | Optional | License number or global identifier for the vehicle, e.g. “E898656”. The field name was chosen to align with the `license_plate` field in GTFS-Realtime. It may specify a different global identifier, particularly for non-road vehicle types without license plates, e.g. Maritime Mobile Service Identity (MMSI) for ferries. |
+
+*Note for future-compatibility:* Future TODS versions may support vehicle couplings: specifically, train cars (individual vehicles) that comprise a train set. Such a proposal is described by the [GTFS-VehicleCouplings draft extension](http://bit.ly/gtfs-vehicles).
```
safrazier17, line 168: For this note, is there anything different that we would expect a rail producer to do in order to future proof their feed? Or is this just for general awareness?

docs/spec/index.md
```diff
@@ -149,37 +149,20 @@ Primary Key: `(date, block_id, service_id)`
...
-| `wheelchair_accessible` | Enum | Optional | Wheelchair accessible. <br />0 or empty - no <br />1 - yes |
-| `seating_capacity` | Non-negative Integer | Optional | This number denotes the number of seats dedicated to riders, excluding folding seats. A seat is considered accommodating only one rider in a seated position. |
-| `max_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders that the vehicle can carry. |
-| `wheelchair_capacity` | Non-negative Integer | Optional | This number denotes the maximum number of riders in a wheelchair that the vehicle can carry. |
+*Note for future-compatibility:* `vehicle_id` field may change to *conditionally required* in a future version where either `vehicle_id` OR `vehicle_type_id` MUST be supplied. `vehicle_type_id` would reference `vehicle_types.txt`, describing vehicle type attributes. This field would be used to assign the type of vehicle in the transit fleet if there is no specific vehicle assignment.
```
safrazier17, line 156: Can we write this note more broadly/generically? At this stage, we don't even have a proposal on `vehicle_type`, which means we also don't know definitely how it would be structured if included. I would favor phrasing this along the lines of "`vehicle_id` field may change to *conditionally required* in a future version where assignments may be made to either an individual vehicle OR a grouping of vehicles."

docs/spec/index.md
antrim: General awareness. Not sure how else we could express this or if it makes sense to include.
I think that a rail producer would advocate for this function to be added.
This is also a heads up for any consumer: Future versions of TODS may have vehicle consists.

docs/spec/index.md
antrim: 593c646

docs/spec/index.md
antrim: b3d2bec

docs/spec/index.md
```diff
@@ -139,3 +139,30 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+Primary Key: `(date, block_id, service_id)`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `date` | Date | Required |  |
+| `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | Optional | Identifies a set of *service days* when the trip is scheduled to take place. Note the [https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#term-definitions](GTFS definition) of *service day* is invoked here. Required if `block_id`s are repeated between different `service_id`s. |
+| `block_id`		  | ID referencing `trips.block_id` | Required | Identifies the block. |
+| `vehicle_id` | ID referencing `vehicles.vehicle_id` | Required | Refers to a specific vehicle in the transit fleet. |
+
+Not every trip or block and date combo needs to have a vehicle specified.
```
skyqrose, line 154: outdated reference to trips that can be removed

```suggestion
Not every block and date combo needs to have a vehicle specified.
```

docs/spec/index.md
```diff
@@ -139,3 +139,30 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+*Note for future-compatibility:* `vehicle_id` field may change to conditionally required in a future version where assignments may be made to either an individual vehicle OR a grouping of vehicles. See [GTFS-Vehicles](http://bit.ly/gtfs-vehicles) for how vehicle categories (types) might be incorporated.
+
+### `vehicles.txt`
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. It is *recommended* but not required to match the `vehicle_id` in GTFS-realtime feeds. |
+| `vehicle_label` | Text | Optional | Free text label for a vehicle, e.g. vessel name. |
```
skyqrose, line 165: I think the most common case of a bus number is worth calling out here.

```suggestion
| `vehicle_label` | Text | Optional | Free text label for a vehicle, e.g. bus number or vessel name. |
```

For comparison, the [GTFS-RT docs](https://gtfs.org/documentation/realtime/reference/#message-vehicledescriptor) describe the label as "User visible label, i.e., something that must be shown to the passenger to help identify the correct vehicle." (Though in TODS we don't need the perspective of passenger experience.)

skyqrose (commented):
this is looking much better now

safrazier17:
### VOTING PERIOD OPEN
Voting is open on this PR through 11:59:59 UTC on April 8, 2025. This PR is one of 3 comprising the TODS 2.1 version proposal. The other PRs associated with the TODS 2.1 proposal are #80 and #87.

### WHO IS ELIGIBLE TO VOTE

Issue Working Group members @BTollison, @jeffkessler-keolis, @skyqrose, @jfabi, @safrazier17, @antrim, @hodb are not eligible to vote on this proposal.

### HOW TO VOTE

### OUTCOMES

timon-k:
+1 [Optibus]

docs/spec/index.md
```diff
@@ -139,3 +139,30 @@ Because some events may overlap in time, it may not be possible to choose a sing
 - Events may have gaps between the end time of one event and the start time of the next. e.g. if an operator's layovers aren't represented by an event.
 - `start_time` may equal `end_time` for an event that's a single point in time (such as a report time) without any duration.
 - Recommended sort order: `service_id`, `run_id`, `event_sequence`.
+
+### `vehicle_assignments.txt`
```
evansiroky, line 143: Please add an entry to this file in the `Files` section of this document.

docs/spec/index.md
```diff
@@ -139,3 +139,30 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+| `date` | Date | Required |  |
+| `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | Optional | Identifies a set of *service days* when the trip is scheduled to take place. Note the [https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#term-definitions](GTFS definition) of *service day* is invoked here. Required if `block_id`s are repeated between different `service_id`s. |
+| `block_id`		  | ID referencing `trips.block_id` | Required | Identifies the block. |
+| `vehicle_id` | ID referencing `vehicles.vehicle_id` | Required | Refers to a specific vehicle in the transit fleet. |
+
+Not every block and date combo needs to have a vehicle specified.
+
+*Note for future-compatibility:* `vehicle_id` field may change to conditionally required in a future version where assignments may be made to either an individual vehicle OR a grouping of vehicles. See [GTFS-Vehicles](http://bit.ly/gtfs-vehicles) for how vehicle categories (types) might be incorporated.
+
+### `vehicles.txt`
```
evansiroky, line 158: Please add a link to this file in the `Files` section of this document.

evansiroky (commented):
Caltrans would like to have some more information on this specification modification proposal before voting. We conceptually like the idea of noting vehicle assignments. However, we echo @sbindman's inquiry and concerns into how this could be used in practice. Vehicle assignments can vary widely from day-to-day or even on an intra-day basis.

It would be great if some clarity could be introduced up-front into the use cases that these vehicle assignments are likely to help resolve and to suggest other methods for delivering vehicle assignments when the use cases demand finer-grained detail of vehicle assignments. 

To guess on how this might be helpful, this seems mainly useful for fleet planning and optimization use cases, but in a realtime sense it may not be useful and could be too rigid to support intra-day or even intra-week vehicle assignment variability. And then from a historical perspective (perhaps even realtime??), it seems like the [TIDES trip_perfromed](https://tides-transit.org/main/tables/#trips-performed) could be more appropriate for noting the exact operational vehicle assignments that ended up occurring.

Also, a minor formatting request is to note these new files at the list of files towards the beginning of the document.

docs/spec/index.md
```diff
@@ -139,3 +139,30 @@ Because some events may overlap in time, it may not be possible to choose a sing
...
+
+Primary Key: `vehicle_id`
+
+| Field Name | Type | Required | Description |
+|---|---|---|---|
+| `vehicle_id` | ID, primary key | Required | Defines an ID for a vehicle. It is *recommended* but not required to match the `vehicle_id` in GTFS-realtime feeds. |
+| `vehicle_label` | Text | Optional | Free text label for a vehicle, e.g. bus number or vessel name. |
```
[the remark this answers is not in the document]
timon-k: I'm not very familiar with GTFS-RT, but it sounds logical to add this hint. 

Would it have a direct benefit to recommend this explicitly? The `vehicle_id` is already recommended to match the one from GTFS-RT, what would an additional sync of the `label` be good for?

timon-k:
> [Sarah from Swiftly] Apologies for raising questions so late during the voting period but I am not clear on who will be able to leverage this spec. In my experience, I have not seen agencies know vehicle assignments far enough ahead of time to put them in a data feed, but rather it is more common to figure out vehicle assignments day of. I would love to learn about real-world cases where this would be applicable to understand whether it makes sense to add scope to the spec. Thanks!

@sbindman Our intended use case is transmitting the vehicle assignments for the immediate next days (typically the next
 1-3 days) into a CAD/AVL system. It's not so much about transmitting a long-term assignment plan, I'm not aware of such a use case in reality.

@evansiroky We did not consider historical data, for that it indeed will be better to use [TIDES Trips Performed](https://tides-transit.org/main/tables/#trips-performed). It's about the assignments for the next few days only.

Apart from the use case of getting existing vehicle assignments into a CAD/AVL system, transmitting the assignments for the next day can also be used for charge management systems (battery-electric buses). That system needs to know which vehicles will enter/leave a depot in the next day so that they can plan the charging in the best possible way.

As with the rest of TODS, we did not have real-time use cases in mind here.

evansiroky:
@timon-k thanks for the clarifications that TODS is intended as a "planned" dataset.

I will go ahead and vote a cautious +1 on behalf of Caltrans as I think the concept makes sense, but still wonder about how applicable this is in the field.

I think it would be great to clarify the timing and use cases. The [GTFS specification notes that](https://gtfs.org/documentation/schedule/reference/#:~:text=the%20published%20gtfs%20dataset%20should%20be%20valid%20for%20at%20least%20the%20next%207%20days):

> the published GTFS dataset should be valid for at least the next 7 days

This is somewhat in conflict with what you've mentioned as the `immediate next days (typically the next
1-3 days)` with TODS, but perhaps this is to be expected in whatever context this is being used.

Matthias-Gruetzner:
I am neutral on PRs 85 and 87. In my eyes, vehicle and driver assignments are a different interface and data is updated on a different schedule, but I see that it can be useful.

arthi-weta:
+1 [San Francisco Bay Ferry]

erik-a-jensen:
Looks like trying this from my work computer yesterday didn't work, but +1 [CTA]

safrazier17:
### VOTING PERIOD CLOSED

The voting period for this proposal closed yesterday, April 15 at 11:59:59 UTC. The proposal received the following votes:

Support (3): @timon-k (Optibus), @evansiroky (Cal-ITP), @arthi-weta (WETA)

docs/spec/index.md
antrim: ?

docs/spec/index.md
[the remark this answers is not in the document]
antrim: Hi @paulswartz Realizing we never fully resolved this. Did not add a note about `vehicle_label`, as `vehicle_id` and `license_plate` are both intended to align with GTFS-RT. Is this ok>

docs/spec/index.md
antrim: Closing for now. I don't see any internal TOC with anchor links to files.

docs/spec/index.md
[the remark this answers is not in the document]
antrim: Closing for now

antrim (commented):
@safrazier17 I believe this is ready to be merged

docs/spec/index.md
antrim: I don't see an in-page TOC ?

docs/spec/index.md
```diff
@@ -155,11 +157,36 @@ This file should represent the schedule after holidays, vacations, and other sch
 
 Each run and date combination may appear 0 times in this file (if there's no assigned employee), 1 time, or multiple times (if multiple employees are assigned to the same run on the same date).
 
-Primary Key: `*`
```
skyqrose, line 158: Is this deletion intended to be here? It may have snuck in during a merge.

safrazier17:
yes you're right, will take care of that this a.m.

fredericsimard:
Replicated in #105

PR #87, opened 2025-03-13
employee_run_dates.txt

skyqrose:
~~Here's an alternative version of #81 , that strips out most of the complexity, leaving just `employee_run_dates.txt`. There's no more roster positions. See [this comment](https://github.com/cal-itp/operational-data-standard/pull/81#issuecomment-2719554507).~~

~~I made Service IDs required again. It would work just fine with optional Service IDs like #81 has, but it doesn't change how many lines are required in the data anymore.~~

~~For discussion purposes. #81 is still the primary proposal.~~

# `employee_run_dates.txt`

This is the output of the Rostering working group: A way for TODS to specify which employees are doing which work on which dates, in 1 new file.

A previous discussion (in #81) considered adding more files to be able to represent more details about rosters, but this PR is now our preferred approach.


safrazier17:
### VOTING PERIOD OPEN
Voting is open on this PR through 11:59:59 UTC on April 8, 2025. This PR is one of 3 comprising the TODS 2.1 version proposal. The other PRs associated with the TODS 2.1 proposal are #80 and #85.

### WHO IS ELIGIBLE TO VOTE

Issue Working Group members @BTollison, @jeffkessler-keolis, @skyqrose, @jfabi, @safrazier17, @antrim, @hodb are not eligible to vote on this proposal.

### HOW TO VOTE

### OUTCOMES

timon-k:
+1 [Optibus]

paulswartz:
+1 @mbta, assuming the Lint checks can be addressed.

docs/spec/index.md
```diff
@@ -133,9 +136,28 @@ Events that don't have `trip_id` set may overlap in time with any other events.
...
 - `start_time` may equal `end_time` for an event that's a single point in time (such as a report time) without any duration.
 - Recommended sort order: `service_id`, `run_id`, `event_sequence`.
+
+### `employee_run_dates.txt`
+
+Describes which employees are scheduled to which runs on which dates.
+
+This file should represent the schedule after holidays, vacation, and other scheduled exceptions have been applied.
+
+Primary Key: `*`
```
rachel-swiftly, line 156: The primary key being `*` suggests that multiple employees could be schedule for the same run and service_id on the same day. While we know this situation occurs in reality when there are service problems, is there a case for scheduling this way? We suggest the primary key be `(date, run_id, service_id)`.

docs/spec/examples.md
```diff
@@ -168,3 +168,67 @@ weekday,10000,20,BLOCK-A,deadhead       ,,garage,08:50:00,stop-1,09:00:00
...
+20240703,B,weekday,102
+20240704,C,weekday,101
+20240704,D,weekday,102
+20240705,C,weekday,101
+20240705,D,weekday,102
+20240706,C,weekend,103
+20240706,D,weekend,104
+20240707,A,weekend,103
+20240707,B,weekend,104
+```
```
rachel-swiftly, line 234: Just wanted to note that it looks like all the rows are present twice :) (one copy of all the rows, then another copy)

rachel-swiftly (commented):
Hi, I'm commenting on behalf of Swiftly. We are generally supportive of this proposal!

In addition to our inline comments, we were wondering if the spec could provide guidance on how often employee_run_dates.txt should be updated / how far out dates should be listed. Our expectation is that this would be updated when a pick is performed and would generally list all dates that the pick and GTFS both cover.

docs/spec/index.md
skyqrose: Yeah, I think this is good. My original thought was that if two employees really are doing the same thing all day (maybe like 2 fare collectors who work the same trains), they could share a run number and the producer could skip some rows. But consumers knowing that there's at most one employee per run is more valuable.

I've made the change.

evansiroky:
In a similar response to https://github.com/cal-itp/operational-data-standard/pull/85#pullrequestreview-2751656000, Caltrans would also like some more information provided on what use cases are hoped to be accomplished with this and also how it relates to realtime/historic distribution of the data. Again, operator assignments could change for a run during a day and the current design of this proposed data model does not allow intra-day changes to who is operating a vehicle.

Also, just pointing out that TIDES uses the "[Operator](https://tides-transit.org/main/tables/#operators)" nomenclature instead of "Employee". Perhaps this TODS update could also reconsider the use of "Employee" since there are some fleets that may use volunteer drivers. Or maybe there will be robots that aren't considered employees that drive a vehicle.

docs/spec/index.md
jeffkessler-keolis: I'd request to revert this change for the original reason cited by @skyqrose: on the railroad front, it is common to have multiple employees assigned to the same run.

While some may pre-specify by craft, others include dual-certified personnel whereby two employees work the same run and are required to vary up who works as the engineer vs conductor on each trip. Likewise, other operations include two matched crew members who work identical shifts in the same function.

In practice, I suspect most consumers can treat `(date, run_id, service_id)` as the primary key, but retaining the flexibility for those of us who schedule that way would be ideal.

Somnambulist57:
+1 [Schedule Masters, Inc]

docs/spec/index.md
skyqrose: Ah, I didn't realize that there were schedules like that in practice. Given that this is a real situation, I'll revert it.

skyqrose:
@evansiroky 

The main use case is including information about who is scheduled to operate each trip within operational software. For example, a dispatcher might want to know if the employee currently operating a vehicle was scheduled to be on that trip or is a substitute. And a field official might want to know which operator is scheduled to do a an upcoming trip, so they can make sure the operator is ready, or they can find a substitute if the operator hasn't shown up yet.

There could be other uses, too, from texting employees to remind them what their schedule will be for tomorrow, or data analysts measuring reasons that employees worked later than scheduled. "What's an employee's schedule" is a pretty fundamental question, so I expect this data to support many different uses.

You're right that this is information that is likely to have real time adjustments, and TODS doesn't have any way to represent that. I expect having scheduled information will still be useful. As a comparison, there's many interesting applications and analyses you can do with static GTFS even without GTFS-RT. And consumers may have realtime information about employees from a separate system. For example MBTA gets data about which operators are logged on to our buses, so in some systems we have realtime operator information but not scheduled operator information.

It's up to the producer to decide how up-to-date to keep this schedule. For comparison, some producers issue updated GTFS feeds for construction or detours issued a week in advance, and some producers don't edit GTFS feeds at all once they're published. This feed is intended to represent the schedule, not reality, so consumers must know that there could have been changes made since the feed was updated, no matter how frequently it's updated.

We intentionally chose "employee" instead of "operator" because this data could represent work besides operating a vehicle. Especially for work like fare collectors, who work on a trips like operators do. It's also possible for `run_events.txt` and `employee_run_dates.txt` to represent information about maintenance jobs, yard shifters, field officials, etc. That's not an intended use case but I think it's nice that it's possible, since agencies probably schedule that work from the same system they use for scheduling operators.

evansiroky:
@skyqrose thanks for the clarifications that TODS is intended as a "planned" dataset.

I will go ahead and vote +1 on behalf of Caltrans.

Matthias-Gruetzner:
I am neutral on PRs 85 and 87. In my eyes, vehicle and driver assignments are a different interface and data is updated on a different schedule, but I see that it can be useful.

arthi-weta:
+1 [San Francisco Bay Ferry]

safrazier17:
### VOTING PERIOD CLOSED

The voting period for this proposal closed yesterday, April 15 at 11:59:59 UTC. The proposal received the following votes:

Support (5): @timon-k (Optibus), @paulswartz (MBTA), @Somnambulist57 (Schedule Masters), @evansiroky (Cal-ITP), @arthi-weta (WETA)

PR #94, opened 2025-04-22
Update change policy from quarterly to annual

safrazier17:
updating change policy to reflect at least annual change cycle, plus minor copy edits


leelas-pm (approved):
Looks good to me!

PR #103, opened 2025-07-10
fix: replaced deprecated action with fork

fredericsimard:
... this new fork is actively maintained. The previous one caused errors during execution.


fredericsimard:
@eliasmbd If you don't mind approving... it's a simple change in which plug in is used in the Lint test action, hopefully it will stop the "Some checks were not successful" like you see above.

eliasmbd (approved):
LGTM

PR #150, opened 2026-07-07
fix: Clarify field types for `start_time`, `end_time`

jfabi:
Follow-up to https://github.com/MobilityData/transit-operational-data-standard/issues/148: Clarifies the `start_time` and `end_time` fields to specify that their `Time` type should match the GTFS definition. This pull request is not meant to make any normative changes to the specification, but rather clarify existing fields.


skyqrose (commented):
LGTM

carlfredl (approved):

PR #156, opened 2026-07-17
docs: clarify v2.1.0 conformance requirements

### LINKED ISSUE #152: Clarify remaining v2.1.0 field and conformance ambiguities
While implementing TODS v2.1.0 conformance checks, I found five ambiguities or apparent errata with no existing issue. I am consolidating them to avoid opening a burst of issues and can split them if preferred.

The duplicate `event_sequence` example is omitted because PR #147 fixed it. The undefined `Time` type is tracked in #148.

## 1. Example CSV values contain alignment padding

Some example rows contain spaces around values, e.g. `daily,10000,10,       ,       ,Operator,Report Time,...`: [source](https://github.com/MobilityData/transit-operational-data-standard/blob/4817010d275f196fff698cb813202c300182dab2/docs/en/spec/examples.md#L109-L110).

A literal CSV reader preserves those spaces. Should examples remove padding, or should consumers trim? The validator compares exactly and warns because padded IDs do not match referents.

## 2. Mid-trip field names differ

The table defines `start_mid_trip` and `end_mid_trip`, while related prose also uses `mid_trip_start` and `mid_trip_end`: [source](https://github.com/MobilityData/transit-operational-data-standard/blob/4817010d275f196fff698cb813202c300182dab2/docs/en/spec/index.md#L124-L129).

May the prose be changed to the table names throughout?

## 3. Meaning of `Primary Key: *`

`employee_run_dates.txt` declares `Primary Key: *`, but TODS does not define `*` for TODS-specific files: [source](https://github.com/MobilityData/transit-operational-data-standard/blob/4817010d275f196fff698cb813202c300182dab2/docs/en/spec/index.md#L167-L181).

Does this import the GTFS convention that all fields jointly form the key? The validator currently warns on exact duplicates pending clarification.

## 4. Required fields on newly added Supplement rows

The Supplement algorithm adds unmatched rows in their entirety, then says the result should be valid GTFS subject to limited exceptions: [source](https://github.com/MobilityData/transit-operational-data-standard/blob/4817010d275f196fff698cb813202c300182dab2/docs/en/spec/index.md#L42-L96).

Must a row adding an entity include every GTFS-required field? The validator requires the primary key and delegates full merged-feed conformance to gtfs-validator.

## 5. Scope of `vehicle_assignments.service_id`

The field is required "if `block_id`s are repeated between different `service_id`s": [source](https://github.com/MobilityData/transit-operational-data-standard/blob/4817010d275f196fff698cb813202c300182dab2/docs/en/spec/index.md#L198-L203).

Is this per row (only where that row's block is ambiguous) or global (one repeated block makes every row require `service_id`)? The validator uses the per-row interpretation.

I am happy to submit follow-up wording for changes the Board considers editorial.

ChelseaKR:
## Summary

- remove alignment padding from CSV examples so copied values do not contain unintended spaces
- use the defined `start_mid_trip` and `end_mid_trip` field names throughout the run-events prose
- spell out all four `employee_run_dates.txt` primary-key fields
- state which GTFS Required fields an added Supplement row must provide
- scope the conditional `vehicle_assignments.service_id` requirement to each row's `block_id`

## Issue #152 disposition

This implements each clarification confirmed in #152:

1. **Example whitespace:** removes the padding from every affected CSV example. Spaces that are part of a field value, such as `Report Time`, are unchanged.
2. **Mid-trip field names:** replaces `mid_trip_start` and `mid_trip_end` in prose with the existing table fields, `start_mid_trip` and `end_mid_trip`.
3. **Employee assignment primary key:** replaces `Primary Key: *` with (`date`, `service_id`, `run_id`, `employee_id`), matching GTFS's meaning of `*` while making the constraint explicit.
4. **Added Supplement rows:** states that a row being added must define all fields marked Required in the corresponding GTFS file, plus any Conditionally Required field whose condition applies to that row. Updates and deletes still need only the Primary Key fields.
5. **Vehicle assignment service:** makes `service_id` conditionally required only for a row whose `block_id` occurs under more than one `service_id`.

## Scope

This PR does not change the `Time` definition. That clarification was handled in #150 and is already on `main`.

The changes clarify the intended behavior discussed in #152; they do not add new files or fields.

## Validation

- `make build`
- `git diff --check`

Closes #152.


skyqrose (commented):
LGTM and satisfies all the discussion from #152

ChelseaKR:
@skyqrose thanks for reading this back in July, and apologies for resurfacing it.

Your "LGTM and satisfies all the discussion from #152" came through as a comment rather than an approval, so the PR still shows "Review required" and has sat open since. If you're still happy with it, a formal approval would let it merge. If it needs a different reviewer, point me at whoever that should be and I'll wait on them.

Nothing has changed since your read. It is the five clarifications confirmed in #152, and it closes that issue.

The primary key is the part I would most like to see land. As written, employee_run_dates.txt says `Primary Key: *`, which leaves implementers to infer which fields are meant. Spelling out (`date`, `service_id`, `run_id`, `employee_id`) makes the constraint something a producer can check their feed against.

skyqrose (approved):
I intentionally gave a comment instead of an approval cuz I'm not too involved in the project anymore. I'm just popping my head back in to provide context from back when I was involved. So I didn't want to assume that I have enough authority to approve a change.

But given that everyone else has had plenty of time to comment, sure, let's say that my approval is enough. Here's the github approval.

skyqrose:
Ah, nevermind, turns out I don't have write access to the repo, so GitHub doesn't count my approval, and you do in fact need approval from someone else.

jfabi (approved):

jfabi:
Thanks for the implicit bump here, and thanks to @ChelseaKR for having put together these clarifications!

I'd raised this with the TODS Board at the end of July to see if anyone thought any of the changes (such as the latter one regarding `vehicle_assignments.service_id`) should be considered normative. Having heard no dissent, we can consider this to be a fully non-normative set of changes—I'll go ahead and merge.

